PR #113, opened 2026-03-25
feat: add 1Password vault provider integration

yaniv-golan:
## Summary

- Add `OnePasswordVaultProvider` implementing the existing `VaultProvider` trait with three auth modes: Service Account, CLI (desktop app), and Connect Server
- v1 ships with explicit mappings only (hostname → item_id + vault_id); domain-based URL search deferred to v2
- Concurrent provider lookup via `tokio::JoinSet` with 500ms grace period (Bitwarden preferred)
- SSRF-hardened Connect client: custom DNS resolver, IPv4+IPv6 blocklist, `no_proxy()`, no redirects, configurable allowlist
- Tokens encrypted at rest via `CryptoService` (AES-256-GCM), decrypted in memory per session
- CORS origin allowlist replaces permissive `mirror_request()` 
- CLI mode gated behind `ONEPASSWORD_CLI_MODE=true` env var
- Opt-in `op` CLI in Docker (multi-arch: amd64/arm64)

## New endpoints

- `GET /api/vault/onepassword/info` — capabilities and available modes (unauthenticated)
- `POST /api/vault/onepassword/discover/accounts` — list signed-in 1Password accounts (CLI mode)
- `POST /api/vault/onepassword/discover/vaults` — list vaults with temporary credentials
- `GET /api/vault/{provider}/vaults` — list vaults for paired account
- `GET /api/vault/{provider}/vaults/{vault_id}/items` — list items in a vault
- `GET /api/vault/{provider}/mappings` — list explicit mappings
- `PUT /api/vault/{provider}/mappings` — create/update mapping (validates item exists)
- `DELETE /api/vault/{provider}/mappings/{hostname}` — delete mapping

## Test plan

- [x] 123 unit tests + 5 integration tests pass
- [x] `cargo clippy -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] End-to-end smoke test with live `op` CLI (v2.32.1) against real 1Password vaults
- [x] Verified: discover accounts, discover vaults, pair (CLI mode), status, list items, create mapping, credential injection via CONNECT proxy, delete mapping, disconnect
- [ ] Service Account mode (needs a service account token)
- [ ] Connect Server mode (needs a 1Password Connect deployment)


yaniv-golan:
Hey @guyb1, that's a helpful pushback actually- you're right to question the op CLI dependency.                                                                                                                                                            
                                                                                                                                                                                                                                                              
Let me walk through the constraints so you have the full picture, and where I landed:                                                                                                                                                                                                     
                                                                                                                                                                                                                                                              
The current state: INSTALL_OP_CLI defaults to false in the Dockerfile, but publish.yml passes true, so the published image ships op to all users regardless of whether they use 1Password. That's not great.                                                

Quick fix - separate image tags (onecli:latest without op, onecli:1password with it) would address the "shipping to non-1P users" concern, but doesn't solve the deeper issue of shelling out to a third-party binary per secret resolution.                

Why alternatives are constrained:                                                                                                                                                                                                                           

- 1Password Connect Server REST API - eliminates the CLI, but requires users to self-host two additional containers (connect-api + connect-sync). That's heavier infrastructure than the CLI, not lighter. @johnnyfish flagged this earlier in the thread as a v2 option.
- Official Rust SDK - doesn't exist. The internal Rust core isn't published as a standalone crate.                                                                                                                                                          
- Unofficial Rust FFI crate - too risky for production.                                                                                                                                                                                                     
- Service Account tokens via direct REST API - not possible. SA auth involves SRP + client-side vault decryption, which is why the SDKs exist.                                                                                                              

What does exist: an official JS/TS SDK (@1password/sdk) that resolves op:// references natively via client.secrets.resolve("op://vault/item/field"). No CLI, no filesystem access, no TCC issues.                                                           

Proposed approach: Move secret resolution from the Rust gateway to the Node.js web app layer, following the same pattern as Bitwarden - where the gateway just decrypts and injects stored credentials, and doesn't talk to the vault backend directly.     

The flow would be:                                                                                                                                                                                                                                          
1. User pairs with SA token (validated via JS SDK, stored encrypted - same UX as today)             
2. User adds hostname → op:// mappings (same API)                                                                                                                                                                                                           
3. Web app resolves op:// references via JS SDK, stores encrypted resolved values in DB
4. Gateway reads encrypted values from DB, decrypts, injects - no op needed                                                                                                                                                                                 
5. Background job periodically re-resolves to handle secret rotation                                                                                                                                                                                        

This removes the op binary entirely, uses an officially supported SDK, and keeps the gateway lean. The tradeoff is a freshness lag on rotation (minutes instead of seconds), but the current implementation already caches for 60s, so comparable in practice.                                                                                                                                                                                                                                                   

This is a bigger change than what's in the PR today - wanted to check if this aligns with what you had in mind before going down that path.

guyb1:
Hey @yaniv-golan, thanks again for the detailed breakdown and for thinking through the alternatives!

One thing we want to avoid is storing resolved secrets from vaults in our DB. We treat the vault as the single source of truth ,and keeping those values in our storage adds a breach surface we'd rather not have. The op:// references are safe to store tho.

What about this flow:

1. User pairs by pasting their SA token, stored encrypted in VaultConnection.connectionData
2. User adds hostname → op:// mappings, stored in DB (references only, not secrets)
3. At request time, the gateway calls the web app via an internal endpoint (POST /api/internal/vault/resolve) with { accountId, opRef, and of course, some authentication method of the agent ( i think agent token would be available at this context and could fit) }
4. Web app creates or reuses a cached JS SDK client for that account's SA token, calls client.secrets.resolve(opRef), returns the value
5. Gateway injects it into the request. The resolved value lives in memory only, never written to the DB
6. SDK clients are LRU-cached in the web app process (e.g. keep 50 most recent, evict after 30 min idle)

This removes the op CLI entirely, uses the official JS SDK, and keeps secrets out of our storage. The tradeoff is the gateway now depends on the web app for vault resolution, which is not great.

I think this is still far from optimal and we should probably try to get in touch with someone from the 1Password team to understand what they'd recommend for this kind of integration. But curious what you think of this direction before we commit to it.

Again, thank you for your time!

natevick:
I want to say how excited I am for a 1Password integration! If there is someway I can join in and help move this forward I will.

mbravorus:
it's been a hot minute since this was last discussed, any chance it is still in the pipeline?

guyb1:
@mbravorus @natevick merged 🎉
would love to hear your feedback after you try it out!

PR #125, opened 2026-03-29
feat: add oauth2 secret type with auto-refreshing access tokens

harche:
## Summary

- Adds a new `oauth2` secret type for APIs that use short-lived OAuth2 tokens (e.g. Google Cloud / Vertex AI)
- The stored credential (service account key or refresh token) is used by a background worker to automatically refresh the access token
- The gateway injects the latest access token as `Authorization: Bearer` — agents never see the real credential

## Changes

**Gateway (Rust):**
- `resolve_encrypted_value()` selects the auto-refreshed access token for oauth2 secrets
- `build_injections()` handles oauth2 as `Authorization: Bearer`
- `SecretRow` includes new `encrypted_access_token` column

**Web app (TypeScript):**
- New `oauth2-service.ts`: Google service account JWT creation + token exchange, generic `refresh_token` grant, background refresh worker (60s poll interval, 5-min pre-expiry margin)
- Validation: `oauth2` type enum, `OAuth2SecretMetadata`, `oauth2Config` schema
- Secret service: oauth2 metadata on create, triggers initial token fetch
- Refresh worker starts via Next.js instrumentation hook, HMR-safe via `globalThis`

**Schema:**
- 3 new nullable columns on `secrets`: `encrypted_access_token`, `access_token_expires_at`, `last_refresh_error`
- Prisma migration included

**Testing:**
- Vitest setup for the web app (first TS test infrastructure in the project)
- 19 TypeScript tests: JWT signing/verification, token exchange (mocked fetch), metadata parsing
- 7 Rust unit tests: oauth2 injection, credential resolution (access token vs stored value)
- All 137 tests pass (118 Rust + 19 TypeScript)

## Motivation

OneCLI currently supports static credentials (`anthropic` API keys, `generic` headers). APIs like Google Cloud Vertex AI use OAuth2 tokens that expire every hour. Without native refresh, users must run external token-refresh daemons. This PR makes OneCLI handle the full lifecycle: store the long-lived credential, refresh the short-lived token automatically, inject it transparently.

## Design decisions

- **Refresh happens in the web app**, not the gateway — keeps the Rust gateway simple (no HTTP client for token endpoints) and reuses the existing crypto service
- **Gateway reads `encrypted_access_token`** instead of `encrypted_value` for oauth2 secrets — the stored credential (service account key) is never sent to upstream APIs
- **No explicit cache invalidation** after refresh — the gateway's 60s connect-resolution cache TTL naturally picks up new tokens, well within the 5-minute pre-expiry refresh margin
- **Two providers**: `google` (service account JWT → token exchange) and `generic` (standard `refresh_token` grant)

## Test plan

- [x] Rust: `cargo test` — 118 tests pass (113 unit + 5 integration)
- [x] TypeScript: `vitest run` — 19 tests pass
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy -- -D warnings` — clean
- [x] `tsc --noEmit` — clean
- [x] `prettier --check` — clean
- [x] `prisma format --check` — clean
- [x] Tested with real GCP credentials: refresh_token → access_token → Vertex AI rawPredict → Claude response

### DIFF
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -111,21 +111,21 @@ model Agent {
 }
 
 model Secret {
-  id                    String    @id @default(uuid())
-  accountId             String    @map("account_id")
-  name                  String
-  type                  String // "anthropic", "generic", "oauth2"
-  encryptedValue        String    @map("encrypted_value") // AES-256-GCM encrypted via CryptoService
-  hostPattern           String    @map("host_pattern") // "api.anthropic.com" or "*.example.com"
-  pathPattern           String?   @map("path_pattern") // "/v1/*", null = all paths
-  injectionConfig       Json?     @map("injection_config") // type-specific config (generic: { headerName, valueFormat })
-  metadata              Json? // type-specific metadata (anthropic: { authMode }, oauth2: { provider, scopes, tokenEndpoint })
-  encryptedAccessToken  String?   @map("encrypted_access_token") // oauth2: current access token (auto-refreshed)
-  accessTokenExpiresAt  DateTime? @map("access_token_expires_at") // oauth2: when the access token expires
-  lastRefreshError      String?   @map("last_refresh_error") // oauth2: last refresh error message
-  createdAt             DateTime  @default(now()) @map("created_at")
-  updatedAt             DateTime  @updatedAt @map("updated_at")
-  account               Account   @relation(fields: [accountId], references: [id])
+  id                   String    @id @default(uuid())
+  accountId            String    @map("account_id")
+  name                 String
+  type                 String // "anthropic", "generic", "oauth2"
+  encryptedValue       String    @map("encrypted_value") // AES-256-GCM encrypted via CryptoService
+  hostPattern          String    @map("host_pattern") // "api.anthropic.com" or "*.example.com"
+  pathPattern          String?   @map("path_pattern") // "/v1/*", null = all paths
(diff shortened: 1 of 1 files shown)


harche:
## Relationship to #123

This PR and #123 solve related but different problems:

| | #123 `google_oauth` | #125 `oauth2` |
|---|---|---|
| **Who refreshes tokens** | Container drives the OAuth flow; gateway intercepts and swaps placeholder credentials in the POST body | Server-side background worker refreshes tokens; gateway injects the access token as a header |
| **What the container sees** | Placeholder `refresh_token` / `client_secret` — container must implement OAuth itself | Nothing — container makes normal API calls, gateway transparently injects `Authorization: Bearer` |
| **Gateway change** | New `ReplaceFormField` injection + body buffering/parsing | Reads `encrypted_access_token` column instead of `encrypted_value` |
| **Use case** | Containers with their own OAuth logic (e.g. Claude Code SDK calling `gcloud auth`) | APIs where you want transparent Bearer injection without the container knowing about OAuth (e.g. Vertex AI, generic OAuth2 APIs) |

**These are complementary, not conflicting.** #123 is the right choice when the container's SDK expects to drive the OAuth flow. #125 is the right choice when you want fully transparent credential injection — the container just calls `https://us-east5-aiplatform.googleapis.com/...` and the gateway handles auth.

Both can coexist — they use different secret types (`google_oauth` vs `oauth2`) and different injection mechanisms (form-body vs header). Happy to rebase on top of #123 if it lands first, or coordinate if both are desired.

PR #187, opened 2026-04-15
feat: multi-account connections, credential_not_found and access_restricted

### LINKED ISSUE #180: Support multi-account OAuth connections for Google services
I'd like to connect both my work account and my personal account on the same onecli instance.

Thanks!

guyb1:
## Summary

- **Multi-account connections**: Support multiple connections per app provider (e.g., two GitHub accounts). Per-request disambiguation via `x-onecli-connection-id` header. Drops the one-per-provider unique constraint, adds `label` column auto-populated from metadata.
- **credential_not_found response**: When an agent hits an unknown host with no matching secret or app connection, returns a structured response with a direct link to create a secret for that hostname.
- **access_restricted broadened**: Now covers both secrets and app connections (previously only checked app connections).
- **MITM all authenticated traffic**: Gateway now intercepts all requests from authenticated agents (not just known app hosts), enabling auth error detection for any host.
- **Auth-related 400 inspection**: Scans response bodies of 400 responses for auth error keywords to catch providers that return 400 instead of 401/403.
- **Multi-account UI**: Connection cards, manage access dialog with per-connection controls, reconnect flow with visual indication, case-insensitive duplicate detection in OAuth callback.
- **Injection rule caching**: Caches decrypted app connection injection rules per-connection in CacheStore with TTL-aware expiration.

## Test plan

- [ ] Connect multiple accounts for the same provider (e.g., two GitHub accounts)
- [ ] Verify `x-onecli-connection-id` header disambiguates when multiple connections match
- [ ] Hit an unknown host with an authenticated agent — should get `credential_not_found`
- [ ] Verify `access_restricted` returned when agent lacks permission to existing connection
- [ ] Disconnect a connection and verify UI updates
- [ ] Reconnect a previously disconnected connection

### DIFF
--- a/apps/gateway/src/connect.rs
+++ b/apps/gateway/src/connect.rs
@@ -18,6 +18,13 @@ use crate::policy::{PolicyAction, PolicyRule};
 /// How long to cache resolved connect responses before re-checking.
 const CACHE_TTL_SECS: u64 = 60;
 
+/// Header name for per-request app connection disambiguation (request).
+pub(crate) const CONNECTION_ID_HEADER: &str = "x-onecli-connection-id";
+/// Header name for listing available connections (response).
+pub(crate) const CONNECTIONS_HEADER: &str = "x-onecli-connections";
+/// Agent secret mode that restricts access to explicitly assigned credentials.
+pub(crate) const SECRET_MODE_SELECTIVE: &str = "selective";
+
 // ── Data types ──────────────────────────────────────────────────────────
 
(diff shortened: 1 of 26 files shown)


johnnyfish:
hey @califa! good news, we just opened PR #187 which adds multi-account connections support, along with credential_not_found and access_restricted handling.

https://github.com/onecli/onecli/pull/187

planning to merge it today. would love for you to give it a spin and let us know how it feels with your setup (personal gmail + workspace account). your feedback would be super helpful here :)

PR #194, opened 2026-04-17
feat: add secret injection in query parameters

### LINKED ISSUE #191: Allow secrets in URL parameters
Sometimes for certain providers (cloud,...) secrets are not exchanged through headers but through URL parameters. onecli should support this scheme

jeffmaury:

YES

Extends other secrets to act on url query parameters rather than HTTP headers

Fixes #191

## What is the new behavior?

Update the create service endpoint and update the UI accordingly

onecli

## Additional context

Add any other context or screenshots.

### DIFF
--- a/apps/gateway/src/connect.rs
+++ b/apps/gateway/src/connect.rs
@@ -718,27 +718,48 @@ fn build_injections(
 
         "generic" => {
             let config = injection_config.and_then(|v| v.as_object());
+
+            // Check for header injection
             let header_name = config
                 .and_then(|c| c.get("headerName"))
                 .and_then(|v| v.as_str());
 
-            let Some(header_name) = header_name else {
-                return vec![];
-            };
-
-            let value_format = config
-                .and_then(|c| c.get("valueFormat"))
+            // Check for parameter injection
+            let param_name = config
+                .and_then(|c| c.get("paramName"))
                 .and_then(|v| v.as_str());
 
-            let value = match value_format {
-                Some(fmt) => fmt.replace("{value}", decrypted_value),
-                None => decrypted_value.to_string(),
-            };
+            if let Some(header_name) = header_name {
+                let value_format = config
+                    .and_then(|c| c.get("valueFormat"))
+                    .and_then(|v| v.as_str());
 
-            vec![Injection::SetHeader {
-                name: header_name.to_string(),
-                value,
-            }]
+                let value = match value_format {
+                    Some(fmt) => fmt.replace("{value}", decrypted_value),
+                    None => decrypted_value.to_string(),
+                };
+
+                vec![Injection::SetHeader {
+                    name: header_name.to_string(),
+                    value,
+                }]
+            } else if let Some(param_name) = param_name {
+                let param_format = config
+                    .and_then(|c| c.get("paramFormat"))
+                    .and_then(|v| v.as_str());
+
+                let value = match param_format {
+                    Some(fmt) => fmt.replace("{value}", decrypted_value),
+                    None => decrypted_value.to_string(),
+                };
+
+                vec![Injection::SetParam {
+                    name: param_name.to_string(),
+                    value,
+                }]
+            } else {
+                vec![]
+            }
         }
 
         _ => vec![],
@@ -960,5 +981,47 @@ mod tests {
     fn build_injections_unknown_type() {
         let injections = build_injections("unknown", "value", None);
         assert!(injections.is_empty());
+    }
+
+    // ── build_injections: paramName ─────────────────────────────────────
+
+    #[test]
+    fn build_injections_generic_param_name() {
+        let config = serde_json::json!({ "paramName": "api_key" });
+        let injections = build_injections("generic", "my-secret", Some(&config));
+        assert_eq!(injections.len(), 1);
+        assert_eq!(
+            injections[0],
+            Injection::SetParam {
+                name: "api_key".to_string(),
+                value: "my-secret".to_string(),
+            }
+        );
+    }
+
+    #[test]
+    fn build_injections_generic_param_name_with_format() {
+        let config = serde_json::json!({ "paramName": "token", "paramFormat": "Bearer-{value}" });
+        let injections = build_injections("generic", "my-secret", Some(&config));
+        assert_eq!(injections.len(), 1);
+        assert_eq!(
+            injections[0],
+            Injection::SetParam {
+                name: "token".to_string(),
+                value: "Bearer-my-secret".to_string(),
+            }
+        );
+    }
+
+    #[test]
+    fn build_injections_generic_header_takes_precedence_over_param() {
+        // If both headerName and paramName are present, headerName wins
+        let config = serde_json::json!({
+            "headerName": "Authorization",
+            "paramName": "api_key"
+        });
+        let injections = build_injections("generic", "my-secret", Some(&config));
+        assert_eq!(injections.len(), 1);
+        assert!(matches!(injections[0], Injection::SetHeader { .. }));
     }
 }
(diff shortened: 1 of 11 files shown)


johnnyfish:
@jeffmaury tnx! @guyb1 take a look 🙏

fbricon:
You can find an example of API key passed as query parameter in https://ai.google.dev/api/generate-content?hl=en#text_gen_text_only_prompt-SHELL

```
curl "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash:generateContent?key=$GEMINI_API_KEY" \
    -H 'Content-Type: application/json' \
    -X POST \
    -d '{
      "contents": [{
        "parts":[{"text": "Write a story about a magic backpack."}]
        }]
       }' 2> /dev/null

```

johnnyfish:
@jeffmaury @fbricon thanks for the PR, really appreciate it. apologies for the delay, we've been planning a larger structural change which is why this took a minute to get to. looking into it now.

if you're around, would be great to jump in our discord and chat through it, [https://discord.gg/PSztzsQB3g](https://discord.gg/PSztzsQB3g). happy to do a call ([https://cal.com/onecli](https://cal.com/onecli)) too if easier, just lmk

apps/gateway/src/gateway/forward.rs
```diff
@@ -166,9 +166,10 @@ pub(crate) async fn forward_request(
         None
     };
 
-    // Apply injection rules matching this request path
-    let injection_count = inject::apply_injections(&mut headers, &path, &rules.injection_rules);
```
johnnyfish, line 170: Mutating `path` here means the injected secret (now in the query string) flows downstream into places it shouldn't:

1. **`PendingApproval.path` at line 243**: the approver sees the raw secret in the path. This defeats the purpose of `sanitized_headers` above (lines 153-167),
which exists specifically to keep credentials out of approval metadata.
2. **Every `url = %url` log statement after this point** (lines 175, 257, 264, 280, 289, 308, 366, 377, 387): the secret ends up in plaintext gateway logs.

Pre-PR this was safe because secrets only went into headers, which were never logged. Suggest keeping the original `path`/`url` for approval storage and logging,
and only using the mutated copy when actually building the upstream `reqwest` request at line 299. Telemetry already strips query params at line 328, so the same
discipline needs to apply to logs and approval metadata.

apps/gateway/src/inject.rs
```diff
@@ -114,13 +120,45 @@ pub(crate) fn apply_injections(
...
+                    count += 1;
+                }
             }
         }
     }
 
     count
 }
 
+/// Add or replace a URL query parameter in a path+query string.
```
johnnyfish, line 134: Heads-up on a subtle behavior change: parsing the existing query with `form_urlencoded::parse` and then re-serializing the whole thing round-trips the pre-existing
params too. That can change their on-the-wire encoding (`%20` vs `+`, `~` encoding or not, ordering).

Fine for most APIs, but breaks signature-based ones (AWS SigV4, OAuth1) where the upstream URL must match byte-for-byte what the caller signed. Worth at least a
doc comment noting the limitation; ideally append-or-replace just the target param without round-tripping the rest.

apps/gateway/src/connect.rs
```diff
@@ -718,27 +718,48 @@ fn build_injections(
...
+            // Check for parameter injection
+            let param_name = config
+                .and_then(|c| c.get("paramName"))
                 .and_then(|v| v.as_str());
 
-            let value = match value_format {
-                Some(fmt) => fmt.replace("{value}", decrypted_value),
-                None => decrypted_value.to_string(),
-            };
+            if let Some(header_name) = header_name {
```
johnnyfish, line 738: Header takes precedence over param when both are set, and there's a test asserting it. Defensive, but the UI/validation layer already prevents both being set, and silently dropping `paramName` could mask a config bug if the data ever gets there some other way (manual DB write, future API). Consider logging a `warn!` when both are present, or rejecting the config at validation time.

apps/web/src/lib/validations/secret.ts
```diff
@@ -1,13 +1,38 @@
...
+  paramName: z.string().min(1),
+  paramFormat: z.string().optional(),
+});
+
 const injectionConfigSchema = z
-  .object({
-    headerName: z.string().min(1),
-    valueFormat: z.string().optional(),
-  })
+  .union([headerInjectionSchema, paramInjectionSchema])
```
johnnyfish, line 18: Non-discriminated union: `{headerName: "...", paramName: "..."}` silently matches `headerInjectionSchema` first and drops `paramName`, instead of erroring.
Consider `.strict()` on each variant, or explicitly rejecting if both fields are present, so partial UI state can't quietly persist.

apps/web/src/app/(dashboard)/connections/_components/secret-dialog.tsx
```diff
@@ -403,34 +429,112 @@ export const SecretDialog = ({
...
                         <>
                           <div className="space-y-2">
-                            <Label htmlFor="secret-header">Header name</Label>
-                            <Input
-                              id="secret-header"
-                              placeholder="e.g. Authorization"
-                              value={headerName}
-                              onChange={(e) => setHeaderName(e.target.value)}
-                            />
+                            <Label>Inject as</Label>
```
johnnyfish: Minor: two `<Button>`s used as a toggle lose the keyboard/screen-reader semantics of a real radio. A `RadioGroup` (or shadcn `Tabs` if you want the visual emphasis) would be more accessible and signal "pick one" more clearly.

apps/web/src/app/(dashboard)/connections/_components/secret-card.tsx
```diff
@@ -22,8 +22,10 @@ import { deleteSecret } from "@/lib/actions/secrets";
 import { SecretDialog } from "./secret-dialog";
 
 interface InjectionConfig {
```
johnnyfish: Same in `secret-dialog.tsx`. With all four fields optional, the type allows ambiguous states (both set, neither set) that don't reflect the wire format. A
discriminated union would mirror reality:

```ts
type InjectionConfig =
  | { headerName: string; valueFormat?: string }
  | { paramName: string; paramFormat?: string };
```

Then the `config?.paramName &&` checks become exhaustive instead of defensive.

johnnyfish (changes requested):
@jeffmaury Nice work overall! One blocker on `forward.rs` (secret leaking into approval metadata + logs), plus a few smaller follow-ups inline.

Want me to push fixes on top, or would you rather take them yourself?

jeffmaury:
> @jeffmaury Nice work overall! One blocker on `forward.rs` (secret leaking into approval metadata + logs), plus a few smaller follow-ups inline.
> 
> Want me to push fixes on top, or would you rather take them yourself?

You can push if you want I have huge pile of work on main projects

johnnyfish:
@jeffmaury No worries, I'll push the fixes on top later today. Thanks for the contribution!

johnnyfish (approved):

johnnyfish:
@jeffmaury thanks again for the contribution! 🙏

We shipped a few related follow-ups on top of this:

- Docs: CLI secrets reference now documents query param injection 
([https://www.onecli.sh/docs/cli/onecli-cli#secrets](https://www.onecli.sh/docs/cli/onecli-cli#secrets))

- [onecli-cli](https://github.com/onecli/onecli-cli): added --param-name / --param-format flags 

[https://github.com/onecli/onecli-cli/pull/32](https://github.com/onecli/onecli-cli/pull/32)

Would love to hear if this works the way you were looking for, drop a comment here or come say hi in Discord

PR #224, opened 2026-04-28
feat: add Vertex AI integration with ADC import and token endpoint interception

### LINKED ISSUE #223: Create a new app to authenticate against Google Vertex AI endpoint
We need sandboxed agents using Anthropic models behind the Google Vertex API endpoints. Google authentication / Vertex configuration is usually done through the gcloud CLI, which generates a config json file, containing client id, secret, refresh token and billing project.

Our plan would be to provide the agent with a fake gcloud config file, and have onecli intercept queries to the aiplatform.googleapis.com, handling oauth2.googleapis.com authentication. onecli google vertex access can be configured by uploading the gcloud config file, or manually setting the required properties.

I have a working PoC, so I will open a PR shortly to discuss implementation details.

fbricon:

YES

Add a new Vertex AI app provider that enables Google Cloud AI Platform
access through the gateway. Users can import their gcloud Application
Default Credentials (ADC) file or fill fields manually to connect.

Gateway changes:
- Add suffix-based host matching for *-aiplatform.googleapis.com
- Intercept oauth2.googleapis.com/token to serve cached access tokens
  when client SDKs attempt their own token refresh through the proxy
- Inject x-goog-user-project header from connection metadata
- Auto-refresh tokens via existing Google OAuth refresh flow

Web app changes:
- New credentials_import connection method type with file upload support
- Vertex AI app definition with ADC JSON file import (keyMap parsing)
- BYOC AppConfig persistence for gateway token refresh
- Updated connect flow UI with secret/non-secret field distinction

Vertex AI athentication is not supported yet.
Fixes #223 

## What is the new behavior?
Screenshot 2026-04-28 at 17 29 28
Screenshot 2026-04-28 at 17 29 33
Screenshot 2026-04-28 at 17 57 06

## Additional context

We're integrating onecli in the [kdn](https://github.com/openkaiden/kdn/issues/270) project and this is one of our required features


johnnyfish (commented):
Really nice work here, the overall architecture is solid and the ADC import UX is a great addition. The suffix matching, token interception, and credentialStub on the app definition all make sense. A few things I noticed while going through:

**expires_in hardcoded to 3600 (forward.rs line 138)**

This always tells the SDK the token is valid for a full hour, but the real cached token might be close to expiring. If it expires in 5 minutes the SDK wont try to refresh again and will just hit 401s for the rest of that window. Could we pull the actual expires_at from the connection credentials and compute remaining seconds? Falling back to 3600 if we dont have it is fine.

**host: "" for suffix-only rules (apps.rs line 218)**

Using empty string as a sentinel feels a bit fragile, any code that reads rule.host directly without going through host_rule_matches would silently get wrong behavior. The test already had to special case it with the is_empty check. Not a blocker but an enum like HostPattern::Exact | Suffix would make invalid states unrepresentable. Could be a followup too.

**account: "" in credentialStub (vertex-ai.ts line 129)**

The other stubs all use "onecli-managed" as the sentinel value for every field. Empty string here breaks that convention, our docs say all sentinels use onecli-managed so theyre easy to detect programatically. Should this be onecli-managed as well?

**dynamic import in route.ts (line 81)**

Small nit, this is the only dynamic import in the file. If its avoiding a circular dep a quick comment would help, otherwise a static import at the top would be more consistant.

@fbricon Overall this is in great shape and excited to get vertex support in. What do you think about those points, any of them you disagree with or see differently?

johnnyfish:
Pushed a small commit to update the Vertex AI SVG logo to the official Google Cloud icon. 
Let me know if that workss, happy to address the issue I raised earlier after

fbricon:
I don't know anything about rust (I'm a java/ts guy), so every praise needs to go to Opus 4.6 ;-)

I'll have it address your points

johnnyfish:
Looking forward to it 🙏 and at least it's 4.6 and not 4.7 😉

johnnyfish:
btw which MCP are you using there?

fbricon:
no MCP for this repo. Using [rtk](https://github.com/rtk-ai/rtk) to limit token usage (eg. when calling cargo via bash)

fbricon:
actually, I'm not too sure how to handle `account`. My own gcloud config file contains an empty value, so I don't know how it's supposed to be injected. I need to look into it.

johnnyfish:
No worries about the account field, it's not used for auth so it doesnt matter for the actual flow. I've already pushed a commit to your branch that addresses the review points (credentialStubs as an array with full paths, account sentinel, static import). Also wired up the CLI side to auto-write stubs safely during `onecli run`.

Will share once its all merged so you can test the full thing end to end. Thanks again for the great work here 🙌

fbricon:
> I've already pushed a commit to your branch that addresses the review points

I don't see that one

anyways, I updated the screenshots with the updated logo.

Thanks for your great feedback anyways

johnnyfish:
Just pushed 2 commits that address the review points (you should see them now). The gateway side got the HostPattern enum, token expiry propagation, and the web app side got credentialStubs as an array with full paths, account sentinel fix, and static import.

Before we merge though, wanted to get @guyb1 `s thoughts on a few things I noticed during the review. Not sure if we should cover them in this PR or handle them after:

1. saveAppConfigWithoutDisconnect duplicates upsertAppConfig minus the disconnect logic. If upsertAppConfig later gets audit logging or encryption changes this function wont get them. Works fine now but we should probaly consolidate into one function with a skipDisconnect option.

2. The token interception in forward.rs is hardcoded to oauth2.googleapis.com + /token + POST. Correct for Vertex AI but if another provider needs similar interception we'd need to generalize it.

3. The oauth2.googleapis.com/token host rule on the Vertex AI provider exists to feed data to the interceptor, not for actual proxying. A comment on that rule in apps.rs explaining this coupling would help future readers.

4. When rules are served from cache, token_expires_at is None so the intercepted response falls back to expires_in 3600. This is acceptable because cache TTL is already capped to the tokens remaining lifetime, so a cached token is gauranteed valid for at least the TTL. Worth documenting though.

5. The POST /api/apps/{provider}/connect route doesnt use the withAudit pattern for state changes. This is pre-existing and not introduced by this PR but worth noting for a followup.

None of these are blockers, just want to align on whether any should go in now vs later.

guyb1:
Hey @fbricon, thanks for this!

question- was there a specific reason you went with authorized_user ADC import over service account JSON keys?

Curious about your thinking here!

fbricon:
we're containerizing agent workspaces with kdn, and providing secret and network isolation through onecli companion containers in podman. So essentially, this is about allowing real humans to provide their own config to their agents (claude code, opencode, cursor...), without them seeing the actual values.
We're not addressing *claw like agent loops at this point in time.

guyb1:
Great @fbricon, thanks! I’m going to push a few edits and fixes today, would love for you to take another look before we merge once they’re up.

fbricon:
Sure I'm watching

guyb1:
Hey @fbricon,

Pushed the changes (sorry about the ugly commit names 😅), main thing is service account JSON support alongside your authorized_user flow.
The connect form auto-detects the file type when you import and shows the right fields. Your existing flow should be untouched.

Made also some other cosmetic stuff, not really that important .

Take a look when you get a chance! and thanks again!

fbricon:
tested it and it works fine, nothing broke on my side. Thanks!

guyb1:
great! merged! @fbricon

PR #227, opened 2026-05-01
feat: make app and gateway ports configurable for multi-user hosts

cfis:
The host-side ports `10254` and `10255` are hardcoded in `docker/docker-compose.yml`, so two users on the same machine can't both run OneCLI — the second container fails to bind. Postgres already has `POSTGRES_PORT` for this; app and gateway didn't.

Adds `ONECLI_APP_PORT` (default `10254`) and `ONECLI_GATEWAY_PORT` (default `10255`), matching the existing `${ONECLI_BIND_HOST:-127.0.0.1}` / `${POSTGRES_PORT:-5432}` convention. Container-internal ports are unchanged — only the host side varies.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -29,13 +29,13 @@ services:
       postgres:
         condition: service_healthy
     ports:
-      - "${ONECLI_BIND_HOST:-127.0.0.1}:10254:10254"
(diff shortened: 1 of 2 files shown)


guyb1:
hey @cfis 

thank you for this contribution! merged.

PR #234, opened 2026-05-01
ci: add SBOM crypto audit to block unapproved crypto dependencies

### LINKED ISSUE #235: Proposal: SBOM crypto-dependency audit for CI
## Problem

OneCLI is a credential vault. Its threat model is anchored on a small set of approved crypto components (Rust: ring, rustls, rcgen, jsonwebtoken; JS: node:crypto built-in). Today, nothing in CI prevents a future PR — even an unrelated one — from silently widening that trusted set by pulling in a transitive crypto crate.

Concrete drift scenarios:

- A new HTTP client or Postgres dep pulls in `openssl-sys` as a transitive
- A new feature reaches for `bcrypt` / `argon2` / `crypto-js` directly
- A maintainer-approved dep upgrade flips a transitive from `ring` to `boringssl` between minor versions

Today, none of these would be caught by automated review. They are only visible to a reviewer who manually inspects `Cargo.lock` and `pnpm-lock.yaml` diffs and recognizes a crypto-relevant package by name.

For a credential vault, this is the kind of thing that is cheap to guard against and expensive to discover after the fact.

## Proposal

Add a small CI guard:

1. A shell script (`scripts/sbom-crypto-audit.sh`) that walks the dep tree on both sides (`cargo tree`, `pnpm list --depth=Infinity`) and exits non-zero if any blocked crypto package appears.
2. A `SBOM Crypto Audit` CI job that runs the script on every PR.
3. On success, the script prints the full approved-crypto inventory so reviewers can see what's actually in the dep tree.

The intent is a regression guard, not a refactor. The script should pass on `main` today (verified locally on `c24a7ac`).

## Open questions for maintainers

- **Block-list scope on JS side**: I'd block `bcrypt` / `argon2` / `scrypt` by default since `node:crypto` covers OneCLI's current needs. If user-password hashing is on the roadmap, those should be approved up front so the guard doesn't bite the future PR.
- **Approve list philosophy**: do you want this hand-curated (like the draft below) or generated from a manifest file? Hand-curated is simpler for a small surface; manifest is better if it grows.
- **Job placement**: standalone `security` job (needs both Rust + Node toolchains) vs. folded into existing `lint` (would need Rust added there). My draft uses standalone — open to either.

## Draft implementation

PR #234 has a working draft if you'd prefer to evaluate the concrete shape rather than discuss in the abstract — happy to close and re-scope based on what you decide here.

## Why I'm proposing this

I run a fork that depends on OneCLI in security-sensitive infra and want the upstream guard rather than maintaining it downstream forever. Hoping to land a few small security/observability improvements (this is the first; docs for the AES-256-GCM wire format and a Node.js↔Rust cross-validation test fixture are next, in separate PRs).

hisgarden:
Refs #235.

> First-time contributor note: CI is currently `BLOCKED` because this PR modifies `.github/workflows/ci.yml`. GitHub doesn't run workflow changes from forked PRs without maintainer approval (sensible default). Happy to demonstrate the new job running green on a self-PR within my fork if that helps before approval.

YES.

Process note: CONTRIBUTING.md asks contributors to open an issue first for new features. Issue #235 is now open with the proposal and open questions. This PR can be reviewed as the concrete draft, or closed and re-scoped from the discussion in #235 — whichever maintainers prefer.

CI / security tooling. New CI job + shell script. No application code is modified.

No automated check on the project's crypto dependency tree. A future PR could pull in `openssl-sys`, `crypto-js`, `node-forge`, `bcrypt`, etc. — directly or transitively — and silently expand the trusted code base of a credential vault. The current safety net is manual review.

Related issue: #235.

## What is the new behavior?

A new `SBOM Crypto Audit` job runs on every PR:

1. Walks the Rust dep tree (`cargo tree`); fails on any blocked crate (`openssl`, `openssl-sys`, `boringssl`, `boring`, `boring-sys`).
2. Walks the JS dep tree (`pnpm list --depth=Infinity`); fails on any blocked package (`crypto-js`, `sjcl`, `node-forge`, `tweetnacl`, `libsodium*`, `bcrypt*`, `argon2`, `scrypt*`).
3. On success, prints the approved-crypto inventory.

**Verification on current `main`:** script passes without any code changes — confirming the existing dep tree is already on the approved list. This PR adds a guard, not a fix.

```
=== SBOM Crypto Audit ===
--- Rust: checking for unapproved crypto crates ---
PASS: No unapproved Rust crypto crates
--- JS: checking for unapproved crypto packages ---
PASS: No unapproved JS crypto packages
AUDIT PASSED: all crypto components are on the approved list
```

## Additional context

**Why this matters for a credential vault**

- Locks the answer to "what crypto is in our trusted base?" into CI.
- Catches supply-chain regressions: a dep upgrade pulling in `openssl-sys` via a new Postgres/HTTP client becomes a CI failure, not a silent expansion.
- Surfaces the trusted set explicitly for security reviewers and downstream packagers.

**Scope / limitations**

- Approve/block lists live in the script — easy to amend.
- Open to discussion on `bcrypt` / `argon2` being blocked: currently blocked because `node:crypto` covers what the project needs, but if user-password hashing is added later those would need to be approved.
- Job adds ~30s to CI after cache warms up.
- New job (not folded into `lint`) because `lint` lacks the Rust toolchain. Happy to relocate if preferred.

**Local CI checks per CONTRIBUTING.md**

`bun run build` and `bun run check` are not relevant to this change (no application code, no types). The script itself was run against this branch on top of `main`; output above.

**Follow-up PRs (independent, opening only after this one's direction is settled)**

- `docs: AES-256-GCM wire format` — documents the at-rest encryption format
- `test(gateway): Node.js cross-validation fixture` — adds a fixed test vector to catch wire-format drift between the TS encrypter and Rust decrypter

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,3 +107,24 @@ jobs:
             --exit-code; EXIT_CODE=$?
           docker rm -f shadow-db
           exit $EXIT_CODE
+
+  security:
+    name: SBOM Crypto Audit
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Setup Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache Rust
+        uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: apps/gateway -> target
+
+      - name: Setup Node.js
+        uses: ./.github/actions/setup-node
+
+      - name: Audit crypto dependencies
+        run: ./scripts/sbom-crypto-audit.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,16 +115,31 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
+      - name: Check for changes
+        id: changes
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            deps:
+              - 'apps/gateway/Cargo.lock'
+              - 'pnpm-lock.yaml'
+              - 'scripts/sbom-crypto-audit.sh'
+              - '.github/workflows/ci.yml'
+
       - name: Setup Rust
+        if: steps.changes.outputs.deps == 'true'
         uses: dtolnay/rust-toolchain@stable
 
       - name: Cache Rust
+        if: steps.changes.outputs.deps == 'true'
         uses: Swatinem/rust-cache@v2
         with:
           workspaces: apps/gateway -> target
 
       - name: Setup Node.js
+        if: steps.changes.outputs.deps == 'true'
         uses: ./.github/actions/setup-node
 
       - name: Audit crypto dependencies
+        if: steps.changes.outputs.deps == 'true'
         run: ./scripts/sbom-crypto-audit.sh
--- a/scripts/sbom-crypto-audit.sh
+++ b/scripts/sbom-crypto-audit.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+################################################################################
+# SBOM Crypto Audit — verify no unapproved crypto libraries in dependency tree.
+#
+# Approved crypto components:
+#   Rust:  ring, rustls, rcgen (cert gen), jsonwebtoken
+#   Rust (transitive, approved): sha2, hmac, cipher, crypto-common, sha1, sha3,
+#          ed25519-dalek, clatter (via Bitwarden Agent Access SDK / sqlx)
+#   JS:    node:crypto (built-in) — no third-party crypto packages
+#
+# Unapproved (would fail this check):
+#   Rust:  openssl, openssl-sys, boringssl
+#   JS:    crypto-js, sjcl, node-forge, tweetnacl, libsodium, bcrypt, argon2, scrypt
+#
+# Usage: ./scripts/sbom-crypto-audit.sh
+################################################################################
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PROJECT_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+ERRORS=0
+
+echo "=== SBOM Crypto Audit ==="
+echo ""
+
+# ── Rust crate audit ──────────────────────────────────────────────────
+echo "--- Rust: checking for unapproved crypto crates ---"
+
+BLOCKED_RUST_CRATES="openssl|openssl-sys|boringssl|boring|boring-sys"
+
+cd "$PROJECT_ROOT/apps/gateway"
+RUST_HITS=$(cargo tree -f '{p}' 2>/dev/null | grep -iE "^($BLOCKED_RUST_CRATES) " || true)
+
+if [ -n "$RUST_HITS" ]; then
+  echo "FAIL: Unapproved Rust crypto crates found:"
+  echo "$RUST_HITS"
+  ERRORS=$((ERRORS + 1))
+else
+  echo "PASS: No unapproved Rust crypto crates"
+fi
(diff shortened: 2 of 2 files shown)


johnnyfish (commented):
Thanks for the well-scoped proposal, @hisgarden - the concept is solid and this is exactly the kind of guard a credential vault should have. Two changes we'd like before merging:

### 1. Add a path filter so the job doesn't run on every PR

The existing `rust` job uses `dorny/paths-filter` to skip when gateway files haven't changed. This job should do the same - only trigger when lockfiles or the script itself change. Something like:

```yaml
- name: Check for changes
  id: changes
  uses: dorny/paths-filter@v3
  with:
    filters: |
      deps:
        - 'apps/gateway/Cargo.lock'
        - 'pnpm-lock.yaml'
        - 'scripts/sbom-crypto-audit.sh'
        - '.github/workflows/ci.yml'
```

Then gate the subsequent steps with `if: steps.changes.outputs.deps == 'true'`.

### 2. Drop the changelog comment at the bottom of the script

The `# Changelog: 2026-03-24 Initial creation` block at the end of `sbom-crypto-audit.sh` - git log already tracks this, no need to duplicate it in the file.

---

Everything else looks good. Clean implementation, passes on current `main`, low maintenance cost. Looking forward to the follow-up PRs you mentioned (wire format docs, cross-validation fixtures).

@guyb1 will also take a look and review.

hisgarden:
Thanks for the review @johnnyfish — both addressed in cd1aa23:

1. **Path filter** — added `dorny/paths-filter@v3` step to the `security` job, gated `Setup Rust`, `Cache Rust`, `Setup Node.js`, and the audit step on `steps.changes.outputs.deps == 'true'`. Filter watches `apps/gateway/Cargo.lock`, `pnpm-lock.yaml`, `scripts/sbom-crypto-audit.sh`, and `.github/workflows/ci.yml` — same pattern as the existing `rust` job.
2. **Changelog block** — removed from the bottom of `scripts/sbom-crypto-audit.sh`.

Ready for another look whenever you and @guyb1 have a moment. Follow-up PRs (wire format docs, cross-validation fixtures) will go up separately so they don't bloat this one.

guyb1:
Thanks for the contribution @hisgarden, the intent is right and we do want this guard in CI.

Before we merge, I think the implementation could be simpler on both sides:

For the Rust side, cargo tree | grep is brittle since it's parsing text output and can hit substring false positives. [cargo-deny](https://github.com/embarkstudios/cargo-deny) handles this properly, it works off Cargo metadata directly and lets you specify blocked crates with use-instead suggestions. Embark Studios built it for banning OpenSSL in favor of rustls/ring, which is literally our case. A deny.toml in apps/gateway/ would replace the shell script, and there's already a GitHub Action for it ([EmbarkStudios/cargo-deny-action@v2](https://github.com/embarkStudios/cargo-deny-action)).

For the JS side, same idea: pnpm list | grep can break across pnpm versions. Parsing pnpm-lock.yaml directly would be more reliable since it's structured YAML. We could also add ESLint no-restricted-imports rules to catch things earlier.

I would love to hear what do you think about it!
Also, want to rework along these lines? Happy to help scope it out.

hisgarden:
Thanks @guyb1.

**Rust:** Agreed — switching to `cargo-deny` with a `deny.toml` (`bans` only for now) and `EmbarkStudios/cargo-deny-action@v2`.

**JS:** I'd lean toward `pnpm list --json --depth=Infinity | jq` over parsing the lockfile directly — same structural robustness, less coupling to pnpm's YAML shape across majors. Happy to go either way; preference?

**ESLint `no-restricted-imports`:** Worth adding as a follow-up — it catches direct imports but not the transitive case this PR is targeting, so I'd keep it separate.

Easier to close #234 and open a fresh PR with the new shape — OK?

PR #259, opened 2026-05-08
fix(gateway): forward absolute-form https:// proxy requests over TLS

Tij8i:
## Summary

The gateway only treats absolute-form `http://` URIs as proxy requests; absolute-form `https://` URIs fall through to the fallback handler and return 400 *connection error: serving HTTP connection*. Most clients use CONNECT for HTTPS, but **axios v1.x with `HTTPS_PROXY` set** doesn't — it configures Node's http/https core agents (used by axios) directly, sending absolute-form `https://` straight to the proxy port. This silently bypasses credential injection for any axios-based MCP server.

This change extends `is_http_proxy_request` to match `https` as well, extracts the inbound scheme in `handle_http_proxy`, and forwards over the matching protocol — reqwest handles TLS transparently.

Per @johnnyfish's invitation in qwibitai/nanoclaw#2330 (cc @guyb1).

## Changes (`apps/gateway/src/gateway.rs`)

1. **`is_http_proxy_request`** — match `Some("http" | "https")` instead of only `"http"`. Docstring updated to call out the axios case.
2. **`handle_http_proxy`** — extract `let scheme: &'static str = match req.uri().scheme_str() { Some("https") => "https", _ => "http" };` alongside the existing authority extraction. (Static-mapped to avoid borrowing from `req`, which is moved into `forward_request` below.)
3. **`forward_request` call** — pass `scheme` instead of the hardcoded `"http"` literal.

## Tests

- `http_proxy_detected_for_https_absolute_uri` — flipped from the previous `not_detected_for_https_uri` (which asserted the old wrong behavior).
- `http_proxy_not_detected_for_other_schemes` — added; locks the http/https-only invariant against `ws://`/etc.
- All 234 unit tests + 7 integration tests pass.

## Local verification

```
cargo fmt --check         # ok
cargo clippy -- -D warnings  # ok
cargo test                # 234 passed, 0 failed (lib) + 7 passed (integration)
```

## Test plan

- [ ] CI green on this branch (cargo fmt, clippy `-D warnings`, cargo test)
- [ ] In a downstream nanoclaw container with the vendored axios monkey-patch removed, `axios.get('https://api.notion.com/v1/users/me', ...)` through the gateway returns 200 with `injections_applied=1` in the gateway log
- [ ] Existing `http://` absolute-form proxy requests continue to work (covered by the existing `http_proxy_detected_for_absolute_uri` test, but worth a smoke check)
- [ ] CONNECT path is unchanged (no edits touch it)

## Source PR

Original debugging + the client-side workaround that motivated this fix: qwibitai/nanoclaw#2330. Once this lands and ships, that workaround can be removed.

### DIFF
--- a/apps/gateway/src/gateway.rs
+++ b/apps/gateway/src/gateway.rs
@@ -405,11 +405,8 @@ async fn fallback() -> StatusCode {
 
 /// An HTTP proxy request has an absolute URI with `http://` or `https://`
 /// scheme (RFC 7230 §5.3.2). Direct requests use origin-form (`/path`).
-///
-/// Most clients use CONNECT for HTTPS, but some (notably axios v1.x with
-/// `HTTPS_PROXY` set) send absolute-form `https://` URIs straight to the
-/// proxy port. We accept both and let `handle_http_proxy` forward upstream
-/// over the matching scheme.
+/// Also matches `https://` because some clients (axios v1.x) send absolute-form
+/// HTTPS URIs to the proxy port instead of using CONNECT.
 fn is_http_proxy_request<T>(req: &Request<T>) -> bool {
     matches!(req.uri().scheme_str(), Some("http" | "https"))
 }
@@ -684,6 +681,7 @@ async fn handle_http_proxy(
     info!(
         peer = %peer_addr,
         host = %authority,
+        scheme = %scheme,
         injection_count = resolved.injection_rules.len(),
         policy_count = resolved.policy_rules.len(),
         "HTTP_PROXY"
@@ -708,11 +706,18 @@ async fn handle_http_proxy(
         rewrite_host: None,
     };
 
+    let http_client =
+        if scheme == "https" && host_matches_skip_verify(&hostname, &state.skip_verify_hosts) {
+            state.http_client_no_verify.clone()
+        } else {
+            state.http_client.clone()
+        };
+
     let mut resp = forward::forward_request(
         req,
         &authority,
         scheme,
-        state.http_client.clone(),
+        http_client,
         &rules,
         &*state.cache,
         &proxy_ctx,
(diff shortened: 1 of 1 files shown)


guyb1:
Thanks for the contribution @Tij8i  ! I merged it with a couple of small fixes (added skip_verify_hosts handling for the https path and scheme to the log line). Mind giving it a quick smoke test to confirm everything still works on your end?

Tij8i:
Thanks @guyb1 — pulled `v1.23.0` and exercised the absolute-form HTTPS path with axios v1.8.4 (same pin as `@notionhq/notion-mcp-server@2.2.1`, which is what originally surfaced this for me).

**The gateway-side fix is working** — `scheme=https` is detected and credential injection runs:
```
INFO onecli_gateway::gateway: HTTP_PROXY peer=172.18.0.1:55388 host=api.notion.com:10255 scheme=https injection_count=1 policy_count=0
```

Hand-crafted absolute-form requests over a raw socket all forward cleanly:

| Request line | Gateway log | Result |
|---|---|---|
| `GET https://api.notion.com/v1/users/me` | `host=api.notion.com` | **200 OK** |
| `GET https://api.notion.com:443/v1/users/me` | `host=api.notion.com:443` | **200 OK** |

**However, axios still fails end-to-end** — the failure mode has just shifted from `400 serving HTTP connection` to a forwarding timeout. Capturing what axios actually writes on the wire (socket-level patch):

```
GET https://api.notion.com:10255/v1/users/me HTTP/1.1
host: api.notion.com
proxy-authorization: Basic ...
```

Axios v1.8.4 stuffs the **proxy's own port (`10255`)** into the absolute-form URI authority. The `Host` header is correct. The gateway honors `req.uri().authority()`, so `forward_request` dials `api.notion.com:10255` — predictably times out 15s later with `WARN connection error error=serving HTTP connection`.

So #259 strictly does what it says (gateway forwards absolute-form HTTPS over TLS); the remaining gap is an axios quirk in URI construction.

If you'd like a defensive mitigation on the gateway side, two options:

1. **Targeted**: if absolute-form authority's port equals the gateway's own listen port (and scheme is https), substitute it with the `Host` header (or 443).
2. **General**: when URI authority and `Host` header disagree, prefer `Host` for the upstream dial.

Happy to test any follow-up. For now we're keeping our container preload that forces `https-proxy-agent` / CONNECT — sidesteps the axios bug entirely.

PR #336, opened 2026-06-01
fix: use documents scope to allow editing existing Google Docs

### LINKED ISSUE #335: Google Docs permissions prevents editing existing files
This is equivalent to #289 for Google Docs rather than Google Sheets.

The existing Google Docs [permissions](https://github.com/onecli/onecli/blob/main/packages/api/src/apps/google-docs.ts#L22) are `drive.readonly` and `drive.file`. For parity with current Google Sheets permissions (see fix in #319), `drive.file` should probably be replaced with `documents`.

levon003:
Summary text below written by Opus 4.8, but I've carefully reviewed the text and agree with Opus that this is likely a change you want to incorporate.

## Problem

The `google-docs` app requests `drive.readonly` + `drive.file`, which only permits editing documents the app itself created/opened. Editing a **pre-existing** Google Doc via `documents.batchUpdate` fails with `403 PERMISSION_DENIED`, even when the connected account has edit rights on the file.

Fixes #335.

## Fix

Mirror the Sheets fix in #319 and its follow-up scope tightening in #323 — replace the broad Drive scopes with resource-specific Docs scopes:

- `drive.readonly` → `documents.readonly` (read)
- `drive.file` → `documents` (read/write — allows editing any existing doc)

This matches the current `google-sheets` definition (`spreadsheets.readonly` + `spreadsheets`) and follows least privilege: the Docs app no longer requests read access to *all* Drive files — Drive listing/search remains the `google-drive` app's responsibility.

## Verification

- `pnpm --filter @onecli/api check-types` ✓
- `pnpm --filter @onecli/api lint` ✓
- `prettier --check` ✓

Verified empirically against the live Docs API: the equivalent Sheets change (#319) took editing an existing spreadsheet from `403` to `200`; Docs is the structurally identical provider.

## Note for existing connections

Existing `google-docs` connections must be reconnected to pick up the new scope (consent re-prompt) — same as Sheets after #319.

### DIFF
--- a/packages/api/src/apps/google-docs.ts
+++ b/packages/api/src/apps/google-docs.ts
@@ -16,20 +16,20 @@ export const googleDocs: AppDefinition = {
       "openid",
       "email",
       "profile",
-      "https://www.googleapis.com/auth/drive.readonly",
-      "https://www.googleapis.com/auth/drive.file",
+      "https://www.googleapis.com/auth/documents.readonly",
+      "https://www.googleapis.com/auth/documents",
     ],
     permissions: [
       {
-        scope: "https://www.googleapis.com/auth/drive.readonly",
+        scope: "https://www.googleapis.com/auth/documents.readonly",
         name: "Read documents",
         description: "View all your Google Docs",
         access: "read",
       },
       {
-        scope: "https://www.googleapis.com/auth/drive.file",
-        name: "Manage app documents",
-        description: "Create and edit documents opened or created by OneCLI",
+        scope: "https://www.googleapis.com/auth/documents",
+        name: "Edit documents",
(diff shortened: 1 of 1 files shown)


explorerleslie:
We just hit this exact limitation in real use: with the current `drive.file` scope, the google-docs app cannot edit any document the user created outside OneCLI — the Docs API rejects writes to anything not created/opened through the app, so a headless agent is effectively read-only on the user's existing documents.

This PR is the same fix that already merged for Sheets in #319 (`spreadsheets` scope). Any chance it could get a review? Happy to test a pre-release build — we're currently deciding between waiting for this or running a locally patched image.

PR #379, opened 2026-06-21
fix(web): redirect /settings at the routing layer to avoid React #310

nbmorgan:

YES

Bug fix.

Navigating to `/settings` white-screens in production with a client-side exception — **Minified React error #310** ("Rendered more hooks than during the previous render").

`/settings` has no page of its own: `app/(dashboard)/settings/page.tsx` calls an **in-render server `redirect()`** to the first settings section. `redirect()` throws `NEXT_REDIRECT` during render, which trips a hook-count mismatch in Next's internal **AppRouter** (de-minified, the throwing `useMemo` is AppRouter's `{ searchParams, pathname }` computation) during client-side **soft-navigation**. It's the only dashboard route doing an in-render redirect, which is why only `/settings` crashed.

## What is the new behavior?

Move the redirect to the **routing layer** via `next.config` `redirects()` (OSS-only; cloud namespaces settings under `/org/<id>`):

```js
async redirects() {
  return isCloud ? [] : [
    { source: "/settings", destination: "/settings/instance", permanent: false },
  ];
}
```

`/settings` now redirects before any page renders, so the in-render `redirect()` is bypassed entirely. Verified: `/settings` no longer white-screens on soft-navigation, and normal navigation (hard load 307 + soft-nav) lands on `/settings/instance` as before.

## Additional context

Found and confirmed on a self-hosted deployment. Affects any OSS/self-hosted user who soft-navigates to `/settings`.

### DIFF
--- a/apps/web/next.config.js
+++ b/apps/web/next.config.js
@@ -28,6 +28,22 @@ const getOssDashboardSegments = () => {
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: "standalone",
+  // `/settings` has no page of its own — it redirects to the first settings
+  // section. Do it at the routing layer instead of an in-render server
+  // `redirect()` in the page, which throws during render and can trip a React
+  // hook-count mismatch (#310) in Next's AppRouter on client soft-navigation.
+  // Cloud namespaces settings under /org/<id>, so this is OSS-only.
+  async redirects() {
+    return isCloud
+      ? []
+      : [
+          {
+            source: "/settings",
+            destination: "/settings/instance",
+            permanent: false,
+          },
+        ];
+  },
   poweredByHeader: false,
   compress: !isCloud, // Cloud: CloudFront handles compression at the edge; OSS: Next.js compresses
   serverExternalPackages: ["@onecli/db", "@1password/sdk"],


mrrobertkent:
Confirming this still reproduces on v1.45.0 (`ghcr.io/onecli/onecli:latest`, OSS edition, `AUTH_MODE=local`), and flagging that the scope has grown since this PR was opened.

### Two more routes now have the same construct

The description notes:

> It's the only dashboard route doing an in-render redirect, which is why only `/settings` crashed.

That held in June, but 1.45.0 added two more redirect-only pages under `(dashboard)`:

- `apps/web/src/app/(dashboard)/connections/apps/page.tsx` → `/connections`
- `apps/web/src/app/(dashboard)/connections/(tabs)/secrets/page.tsx` → `/connections/custom`

Both are the same shape as `settings/page.tsx` — a server page whose entire body is an in-render `redirect()`:

```js
export default function AppsRedirectPage() {
  redirect("/connections");
}
```

As written, the `redirects()` entry in this PR resolves `/settings` only, so those two would keep white-screening on soft-navigation after it merges. Adding them alongside the existing entry looks sufficient:

```js
{ source: "/connections/apps",    destination: "/connections",        permanent: false },
{ source: "/connections/secrets", destination: "/connections/custom", permanent: false },
```

`/connections/secrets` reads `searchParams` and forwards a query string, and it branches to `/connections/llms?create=…` when `create` is `anthropic` or `openai`. A static `redirects()` entry preserves neither, so that one may warrant either a `has` condition for the `create` cases or leaving its in-render redirect in place — worth a maintainer's call rather than assuming.

### Reproduction

Against the container on loopback, with no reverse proxy in the request path:

```
$ for p in /settings /connections/apps /connections/secrets /setup-error; do
    curl -s -o /tmp/b -D /tmp/h "http://127.0.0.1:10254$p"
    printf "%-24s %-32s %s\n" "$p" "$(head -1 /tmp/h | tr -d '\r')" \
      "$(grep -o 'NEXT_REDIRECT;[^\\"]*' /tmp/b | head -1)"
  done

/settings                HTTP/1.1 200 OK                  NEXT_REDIRECT;replace;/settings/instance;307;
/connections/apps        HTTP/1.1 200 OK                  NEXT_REDIRECT;replace;/connections;307;
/connections/secrets     HTTP/1.1 200 OK                  NEXT_REDIRECT;replace;/connections/custom;307;
/setup-error             HTTP/1.1 307 Temporary Redirect  <no sentinel>
```

All three return `200 text/html` with the serialized `NEXT_REDIRECT` sentinel in the body instead of an HTTP 307, so the redirect is deferred to the client router. `/setup-error` is a useful control: it also calls `redirect()`, but its redirect is issued by `proxy.ts` middleware before the page renders, and it returns a real 307. That isolates the failure to page-level in-render redirects rather than to `redirect()` generally.

Worth noting the sentinel appears on a plain hard `GET` as well, not only on soft-navigation.

### Workaround for anyone self-hosting meanwhile

Until this lands, the same "resolve before any page renders" strategy can be applied at the reverse proxy. On Traefik, three `redirectRegex` middlewares anchored with `/?$` so sub-routes are untouched:

```yaml
onecli-redir-settings:
  redirectRegex:
    regex: "^(https?://[^/]+)/settings/?$"
    replacement: "${1}/settings/instance"
    permanent: false
```

Verified against all three paths, with `/settings/instance` and `/connections/apps/<id>` confirmed to pass through unchanged.

mrrobertkent:
Correction to my comment above: I said 1.45.0 *added* `/connections/apps` and `/connections/secrets`. That's wrong — `git log --diff-filter=A` puts both at `aa8ec62` (2026-03-29, #127), roughly three months before this PR was opened. I had read `git log -1 -- <file>`, which reports the last commit to touch a file, as the commit that introduced it.

The finding itself is unchanged, but the framing was backwards: these aren't new routes that appeared after the PR was written, they're two pre-existing pages with the same in-render `redirect()` shape that the original scoping missed. Verified against both this branch's base and current `main` — the files are present in each.

Sorry for the noise.

mrrobertkent:
Offering the two-route extension as a branch, if it is useful to you.

I have `/connections/apps` and `/connections/secrets` implemented on top of this branch and running in production. The second needs a little care — it branches on `?create=anthropic|openai` to the LLM tab, so a plain entry would lose that. Expressed as a `has` rule ahead of the catch-all:

```js
{
  source: "/connections/secrets",
  has: [{ type: "query", key: "create", value: "(?<create>anthropic|openai)" }],
  destination: "/connections/llms?create=:create",
  permanent: false,
},
{ source: "/connections/secrets", destination: "/connections/custom", permanent: false },
```

Verified against the running app — all five cases now emit a real 307 from the routing layer:

```
/settings                              307 -> /settings/instance
/connections/apps                      307 -> /connections
/connections/secrets                   307 -> /connections/custom
/connections/secrets?create=anthropic  307 -> /connections/llms?create=anthropic
/connections/secrets?foo=bar           307 -> /connections/custom?foo=bar
```

(The last one confirms Next forwards the original query when the destination declares none, so no explicit handling is needed for the general case.)

Say the word and I will open a PR against `fix/settings-redirect` — it is your fix and your PR, so I would rather offer than push a cross-fork PR at you unasked. Equally happy for you to just take the snippet.

PR #389, opened 2026-06-25
feat(web): show build version on Overview and Settings

### LINKED ISSUE #388: Dashboard doesn't show which OneCLI version is running
### Problem

The dashboard gives no indication of which OneCLI build an instance is running:

- **Overview** and **Settings** show no version anywhere.
- `GET /api/health` returns `"version": "unknown"`.

On a self-hosted instance this makes it hard to:
- confirm a deploy actually picked up new code (vs. serving a cached image),
- match observed behavior to a release when filing a bug report.

### Proposal

Surface the build version in the dashboard and report it from the API:

- Show it on the **Overview** page and as a card on **Settings → Instance**.
- Report it from `GET /api/health` instead of `"unknown"`.
- Resolve the version from the monorepo `package.json`, with an optional
  `APP_VERSION` / `ONECLI_VERSION` build-time override (and a Docker
  `--build-arg`) so custom/tagged images can report their actual tag.

Happy to contribute this — I have a PR ready.

nbmorgan:
Closes #388.

### What

Surfaces the running OneCLI build version in the dashboard and the API:

- **Overview** page: a small version line in the footer.
- **Settings → Instance**: a "Build version" card.
- `GET /api/health`: reports the version instead of `"unknown"`.

### How

- `next.config.js` resolves the version at build time and exposes it as
  `NEXT_PUBLIC_APP_VERSION`: `ONECLI_VERSION` / `APP_VERSION` env override when
  set (e.g. a tagged image), otherwise the monorepo `package.json` version;
  falls back to `"dev"`.
- A small `lib/version.ts` helper (`APP_VERSION`) is read by the UI and passed
  into `createApiApp` for `/api/health`.
- `docker/Dockerfile` takes an optional `--build-arg APP_VERSION` so a built
  image can report its actual tag.
- The two build-affecting env vars are declared in `turbo.json`.

### Notes

No new dependencies. Server- and client-side both read the same injected value,
so the dashboard and `/api/health` always agree.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,6 +45,10 @@ jobs:
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
+      - name: Resolve app version
+        id: appversion
+        run: echo "value=$(jq -r .version package.json)" >> "$GITHUB_OUTPUT"
+
       - name: Build and push by digest
         id: build
         uses: docker/build-push-action@v7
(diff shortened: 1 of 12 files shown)


guyb1:
Great catch, thanks for the PR @nbmorgan! Made a few small tweaks and merged

Issue #192, opened 2026-04-17
Improve GitHub connection: support both OAuth Apps and GitHub Apps


jezekra1:
The current GitHub connection in `apps/web/src/lib/apps/github.ts` only implements the classic OAuth App flow, and mixes GitHub App conventions into it in ways that produce a confusing, broken experience when users try to connect a GitHub App.

## Current limitations

1. **No GitHub App installation step.** `buildAuthUrl` only hits `https://github.com/login/oauth/authorize`. For a GitHub App, the user must first be sent to `https://github.com/apps/<app-slug>/installations/new` (or `/settings/installations/<id>`) to select the account/repos to install on. Without this, a GitHub App connection never becomes usable — no `installation_id`, no repo access.

2. **Permissions shown in the UI don't reflect GitHub Apps.** `permissions: [...]` lists classic OAuth scopes (`repo`, `user`, `gist`, `notifications`, `project`, `codespace`, `workflow`). GitHub Apps use fine-grained, per-resource permissions (e.g. `contents:read`, `issues:write`, `pull_requests:write`, `metadata:read`) declared on the App itself — not OAuth scopes. For a GitHub App connection the UI shows scopes that don't apply, and hides the permissions that do.

3. **Config placeholders imply GitHub App, flow implements OAuth App.** The Client ID placeholder (`Iv1.abc123...`) matches the GitHub App Client ID format, but the token exchange and scope handling are OAuth-App-only. Users pasting a GitHub App's credentials get a half-working connection.

4. **No installation/permission introspection.** For a GitHub App we should resolve `installation_id`, list accessible repos, and render the App's declared permissions from `GET /app` / `GET /user/installations` — so the UI reflects what the App can actually do.

## Proposal

Support both connection types for GitHub as first-class options:

- **OAuth App** (user-to-server token, classic scopes) — current flow, cleaned up.
- **GitHub App** (installation access token, fine-grained permissions):
  - Redirect to the installation URL; handle the `installation_id` / `setup_action` callback.
  - Exchange user-to-server code + track `installation_id` per connection.
  - Fetch the App's declared permissions and installed repositories; render those in the UI instead of OAuth scopes.
  - Handle installation token refresh (JWT-signed, 1h TTL) on the server.

This likely requires extending `AppDefinition` / `connectionMethod` to distinguish the two variants (e.g. `type: "oauth"` vs `type: "github-app"`) rather than forcing GitHub App semantics into the OAuth shape.

jezekra1:
[...]

Issue #181, opened 2026-04-12
Support additional OAuth2/OIDC providers (e.g. via Dex or generic OIDC)


niiku:
Hey, first off - really awesome project. I've been looking for a solution that handles credential injection for agents and OneCLI is the first one I've found that actually does this well. The idea of using good ol' forward proxy is brilliant. Thanks a lot! :-)

I have a couple of improvements in mind. So I will create a couple of issues - hope you don't mind. I can help develop, but of course, first wanted to kick off a bit the discussion to have an aligned understanding how to approach these ideas (if, at all)

Currently the web app only supports Google as an OAuth provider (hardcoded in `nextauth-config.ts`). For self-hosted and enterprise setups it would be really valuable to support other identity providers like GitHub, GitLab, Okta, Entra ID, etc.

A couple of approaches come to mind:

- Dex as an OIDC broker - put something like [Dex](https://dexidp.io/) in front of OneCLI so operators configure their upstream IdPs in Dex, and OneCLI just speaks standard OIDC to a single endpoint. 
- Make the NextAuth provider list configurable - since the app already uses NextAuth (which supports dozens of providers), the provider could be selected via environment variables / config file instead of being hardwired to Google.

What do you think? :-)

glifocat:
I took a stab at this. I have a small patch against v1.41.0 that adds a
generic OIDC provider next to Google, configured with `OIDC_ISSUER`,
`OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`, and an optional `OIDC_PROVIDER_NAME`
for the login button label. Google deployments are unchanged, and the
provider id is `oidc`, so the callback URL is
`<APP_URL>/api/auth/callback/oidc`. I tested the full login flow end to end
against Pocket ID 2.10.0, and the provider goes through standard OIDC
discovery, so any compliant IdP should work, including the Dex route
suggested here.

Opening a draft PR shortly. Happy to adjust the approach if the maintainers
prefer a different shape (for example NextAuth provider config instead of
dedicated env vars).

Assisted by Claude; reviewed and verified by a human before submission.

PR #421, opened 2026-07-08
fix(api): resolve the OAuth callback's app origin from configured APP_URL

### LINKED ISSUE #420: OAuth callback always redirects to localhost:10254 when APP_URL is not set (the fallback never runs)
## What happens

When someone self-hosts OneCLI and does not set the `APP_URL` variable, every OAuth connect flow ends on the wrong page. The provider consent step works. But the last redirect sends the browser to `http://localhost:10254` instead of the real site.

## Why it happens

In `packages/api/src/routes/apps.ts`, the callback picks the redirect target like this:

```ts
const appOrigin = APP_URL || apiOrigin;
```

It looks like it will fall back to the request origin (`apiOrigin`) when `APP_URL` is empty. But `APP_URL` here is the constant from `env.ts`:

```ts
process.env.APP_URL ?? process.env.NEXT_PUBLIC_APP_URL ?? "http://localhost:10254"
```

This constant always has a value. It is never empty. So the `|| apiOrigin` part never runs. It is dead code. When `APP_URL` is not set, `appOrigin` is always `http://localhost:10254`.

The first half of the flow already does this right. The consent `redirect_uri` comes from the request (`lib/request-origin.ts`, added in #325). So the flow half works: consent is fine, then the last hop strands the user on localhost.

## How to see it

1. Self-host the gateway on any host that is not localhost. Do not set `APP_URL`.
2. Connect any OAuth provider from the dashboard.
3. Consent works. Then the browser lands on `http://localhost:10254/app-connect/...`

I saw this live on a headless install of v1.40.0 (`61f6238`). The same code is on main today (`fe0981f`, checked 2026-07-08, `apps.ts` line 435).

## Suggested fix (about 2 lines)

In the callback, read the raw env values the same way `lib/request-origin.ts` does:

```ts
process.env.APP_URL ?? process.env.NEXT_PUBLIC_APP_URL
```

This is `undefined` when not set. Then the existing `|| apiOrigin` fallback starts working. The newer request-origin code (after #407) already uses this raw-env pattern, so the fix matches what the project already does.

## Related, but not blocking

The links agents hand out (`lib/skills/gateway-skill.ts` through `lib/dashboard-url.ts`) use the same constant, so they have the same localhost default. There is no request in scope there, so that one needs its own design call. I mention it here so it is not forgotten.

---

Assisted by Claude; reviewed and verified by a human before submission.

glifocat:

YES

Bug fix. Fixes #420.

---

> **Maintainer note.** @glifocat's diagnosis here was exactly right, and the fix proposed in the **issue** (#420) was the correct design — read the raw env so it is `undefined` when unset, and let the existing `|| apiOrigin` fallback finally run. That is what this PR now implements.
>
> The version that had been pushed to the branch simplified it further, to `const appOrigin = apiOrigin`, and that turned out to break deployments that serve the API and the dashboard on **different hosts** (details below). I've reset the branch to `main`, re-applied the fix against current code, and extended it to the other places with the same defect. Credit for finding this is entirely @glifocat's — thanks for the unusually clear write-up, including pinning the exact line and commit.

## The bug

`packages/api/src/routes/apps.ts` picked the callback's redirect target with:

```ts
const appOrigin = APP_URL || apiOrigin;
```

`APP_URL` there is the constant from `lib/env.ts`, which defaults to `http://localhost:10254`. It is never falsy, so `|| apiOrigin` was unreachable dead code. Self-hosters who had not set `APP_URL` completed provider consent and were then sent to localhost. Also reported independently in #192.

## Why not "just use the request origin"

Those are two different questions, and conflating them is what caused this in the first place:

- **Who answered this request?** → `getRequestOrigin()`
- **Where does a browser go for a dashboard page?** → not necessarily the same host

A deployment can run this API package standalone — answering on an API host and declaring it via the `selfUrl` option — while the dashboard is served elsewhere. CLI and SDK clients talk to the API host, so a browser really can arrive at the callback there. But `/app-connect/*` is a Next.js page that only exists on the dashboard host, so the last redirect has to cross origins, and it can only do that from a configured `APP_URL`.

`apps-callback-origin.test.ts` locks that in: with `selfUrl` on one host and `APP_URL` on another, the redirect must land on the dashboard host. It fails with `https://api.example.com/app-connect/...` if the preference is dropped — the literal dead URL a user would hit.

## The fix

- **`lib/app-origin.ts`** (new) — `configuredAppUrl()` returns the explicitly configured URL, or `undefined` when unset/empty/whitespace. That `undefined` is the whole point: it is what lets a call site say "…and otherwise use the request origin". Plus `originFromHeaders()`, the header-derivation logic lifted out of `request-origin.ts`.
- **`getAppOrigin(request)`** in `lib/request-origin.ts` — names the invariant so it is harder to break: configured `APP_URL` first, request origin only as the fallback.
- Applied at the callback and at the agent-facing "go set up your credentials" link (`dashboardUrl` gained an optional fallback origin).
- **Settings → Instance** now shows the origin actually in use, with an auto-detected note. Side benefit: that page was statically prerendered, so it had been displaying its *build-time* `APP_URL`; it is dynamic now.
- `originFromHeaders()` validates the host and scheme it derives. These origins reach `Location` headers and, on the OAuth fragment-bridge path, the inside of a `<script>` block escaped with `JSON.stringify` — which does not neutralize `</script>`.

| Deployment | `APP_URL` | request origin | result |
| --- | --- | --- | --- |
| single host | set | same | unchanged |
| **split API / dashboard hosts** | set | **differs** | **redirect stays on the dashboard host** |
| self-hosted, set | set | = it | unchanged |
| **self-hosted, unset** | unset | derived | **fixed** |

## Not included

The `docker-compose.yml` `APP_URL` pin. Unpinning it is right in principle — deriving a *public* URL from a *bind* address is a category error — but `scripts/install.sh` auto-selects the docker0 bridge IP on bare-metal Linux and publishes only there, so `127.0.0.1:10254` is not listening. The gateway builds absolute links with no request to derive from (`apps/gateway/src/gateway/response.rs`), and unpinning alone would point those at a dead localhost. Those two have to change together; tracked separately.

## Verification

- `@onecli/api`: check-types, lint (`--max-warnings 0`), **715 tests** passing.
- `@onecli/web`: check-types, lint. Prettier clean.
- Each regression test was confirmed to fail against the original bug before being kept.


glifocat:
Checking in on this one. It has been ready for review since July 8. The
companion feature PR #430 (generic OIDC login, from the #181 discussion) is
now ready for review as well, in case both can share one review pass.

I do not see CI runs on either PR yet. I think they are waiting on a
maintainer to approve workflows for a first time contributor. Happy to rebase
or adjust either one if anything looks off.

Assisted by Claude; reviewed and verified by a human before submission.

guyb1:
thanks @glifocat , merged

Issue #236, opened 2026-05-02
OAuth bearer not injected on https://www.googleapis.com/batch/calendar/v3 — Google Calendar batch endpoint silently returns auth errors


tianglim:
## Summary

Outbound requests from agents to `https://www.googleapis.com/batch/calendar/v3` pass through the gateway with `injections_applied=0`, even when the `google-calendar` app is connected with valid OAuth tokens. The same agent hitting `https://www.googleapis.com/calendar/v3/calendars/<id>/events` gets correctly authenticated (`injections_applied=1`).

The batch endpoint is what Google's official `googleapis` client library (and downstream MCP servers like [`@cocal/google-calendar-mcp`](https://github.com/cocal-com/google-calendar-mcp)) uses any time you query multiple calendars in one call. Because the auth header isn't swapped, every sub-request inside the batch returns:

> Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential.

The MCP swallows these errors and returns `events: []`, so the agent believes the calendars are empty and reports "clear day" even when they're full.

## Repro

```bash
# Inside a container with HTTPS_PROXY=onecli's gateway and
# Google Calendar app connected to OneCLI:

curl -s -X POST "https://www.googleapis.com/batch/calendar/v3" \
  -H "Authorization: Bearer onecli-managed" \
  -H "Content-Type: multipart/mixed; boundary=batch_test" \
  --data-binary @batch.txt
```

Where `batch.txt` is a normal Google batch request body. Compare to a direct call to `/calendar/v3/calendars/primary/events` — the latter returns 200 with events, the former returns 200 with multipart sub-bodies each carrying a 401.

## Gateway log evidence

Same agent, same session, two consecutive requests:

```
MITM method=GET  url=https://www.googleapis.com:443/calendar/v3/calendars/primary/events  injections_applied=1
MITM method=POST url=https://www.googleapis.com:443/batch/calendar/v3                    injections_applied=0
```

## Impact

Affects any consumer of Google's batch API while routed through OneCLI — most notably the official `@cocal/google-calendar-mcp` MCP server when queried with multiple calendars (its array form). Users running a multi-calendar setup get a silent "no events" answer that looks correct but is empty.

## Suggested fix

[...]

PR #429, opened 2026-07-15
fix(gateway): merge secret and OAuth app injection on shared hosts

### LINKED ISSUE #428: Gateway: API-key secret injection preempts OAuth app injection on shared hosts (www.googleapis.com)
## Summary

When a host carries **both** a vaulted API-key secret and one or more OAuth **app** connections, the gateway injects only the secret and **skips OAuth app resolution entirely** for that host. OAuth requests then reach the provider with no `Authorization` header and fail upstream with `401`, which the gateway surfaces as `app_not_connected`.

This breaks **every** Google Workspace OAuth app (Calendar, Drive, and the Gmail legacy path) for any project that also has a **YouTube Data API key** vaulted — because all of them live on `www.googleapis.com`. Requests to the providers' **dedicated** hosts (`gmail.googleapis.com`, `tasks.googleapis.com`) are unaffected, which is the tell.

This is distinct from #236/#237 (which added missing `/batch/` path rules): here the path rules are correct and present, but app resolution never runs because a secret is bound to the same host.

## Repro

In a project that has (a) a Google Calendar OAuth app connected **and** (b) a YouTube Data API key vaulted as a secret on `www.googleapis.com` (path `/youtube/`, query param `key`):

```bash
# Agent request through the gateway, no auth header — gateway should inject the Bearer:
curl -s -o /dev/null -w "%{http_code}\n" \
  "https://www.googleapis.com/calendar/v3/users/me/calendarList"
# => 401   (gateway logs: "app not connected ... provider=google-calendar")

# Same OAuth connection, dedicated host — works:
curl -s -o /dev/null -w "%{http_code}\n" \
  "https://gmail.googleapis.com/gmail/v1/users/me/profile"
# => 200
```

Remove the YouTube key (or query the provider before the key is vaulted) and the calendar request returns `200`. So it is the *coexistence* of a secret and an app on the same host that triggers it — the OAuth token itself is valid (it works on the dedicated host).

## Root cause

App-connection resolution is gated on there being **no** secret rules. In `apps/gateway/src/gateway/mitm.rs`:

```rust
let mut injection_rules = resp.injection_rules; // from secrets
// If no secret rules, try app connections (per-request disambiguation)
if injection_rules.is_empty() && !resp.app_connections.is_empty() {
    // ... resolve_app_injection_for_request ...
    injection_rules = rules;   // and: replace, not merge
}
```

The YouTube key makes `injection_rules` non-empty, so the whole block is skipped and `resolve_app_injection_for_request` is never called. The identical pattern is in `apps/gateway/src/gateway.rs` (the HTTP-proxy path).

Secret rules and app rules are both **path-scoped** `InjectionRule`s (`/youtube/*` for the key vs `/calendar/*` for the OAuth Bearer), so they can and should coexist rather than one preempting the other.

## Evidence

Instrumenting the `apply_injections` call site, a calendar request carries only the youtube rule:

```
path=/calendar/v3/... n_rules=1 patterns=[("/youtube/*", 1)] injection_count=0 auth_len=0
```

After the fix (run app resolution regardless and merge), the same request carries both, and the `/calendar/*` Bearer applies:

```
path=/calendar/v3/... n_rules=3 patterns=[("/youtube/*",1),("/calendar/*",1),("/batch/calendar/*",1)] injection_count=1 auth_len=260
```

A `/youtube/*` request still injects the key and correctly ignores the non-matching OAuth rules.

## Proposed fix

Run app resolution whenever app connections exist, and **merge** the resulting rules with any secret rules (`injection_rules.extend(rules)`) instead of gating on `injection_rules.is_empty()` and replacing. Apply to both `mitm.rs` and `gateway.rs`. Reproduced against `main` (32b5852).

Happy to open a PR with the fix + a regression test (I have both ready) if the direction looks right.

## Environment

- OneCLI `main` (32b5852); also reproduced on the released `latest` image.
- Self-hosted, Google BYOC OAuth apps (Gmail/Calendar/Tasks/Drive) + YouTube Data API key as a vaulted secret.

Related: #236 / #237 (batch path rules — different root cause), #307 (token endpoint routing — adjacent).

---

*Disclosure: I found this bug while integrating OneCLI, diagnosed it with an AI coding assistant, and verified the repro, root cause, and fix against a local build myself before filing.*

jbaruch:

YES

Bug fix (gateway). Fixes #428.

When a host carries **both** a vaulted API-key secret and one or more OAuth **app** connections, the gateway injects only the secret and skips OAuth app resolution entirely for that host. OAuth requests then reach the provider with no `Authorization` header and 401, surfaced as `app_not_connected`.

Concretely: with a YouTube Data API key vaulted as a secret on `www.googleapis.com` (`/youtube/`), every Google Workspace OAuth app that shares that host — Calendar (`/calendar/`), Drive (`/drive/`), and the Gmail legacy path (`/gmail/`) — stops authenticating. The same OAuth connection works on its dedicated host (`gmail.googleapis.com`), which is the tell.

Root cause: app-connection resolution is gated on there being no secret rules — `if injection_rules.is_empty() && !resp.app_connections.is_empty()` — in both `apps/gateway/src/gateway/mitm.rs` and `apps/gateway/src/gateway.rs`. The YouTube key makes `injection_rules` non-empty, so `resolve_app_injection_for_request` is never called. This is distinct from #236/#237 (which added missing `/batch/` path rules); here the path rules are correct but app resolution is skipped.

## What is the new behavior?

Run app resolution whenever app connections exist, and **merge** the resulting rules with any secret rules (`injection_rules.extend(rules)`) instead of gating on `injection_rules.is_empty()` and replacing. Secret and app rules are both path-scoped (`/youtube/*` vs `/calendar/*`), so they coexist: a `/calendar/` request now applies the OAuth Bearer, and a `/youtube/` request still applies the API key and ignores the non-matching OAuth rules. Applied to both the MITM and HTTP-proxy paths.

## Additional context

**Verification.** Reproduced and fixed against `main` (32b5852), and verified live end-to-end on a self-hosted gateway. Instrumenting the `apply_injections` call site, before the fix a calendar request carried only the youtube rule:

```
path=/calendar/v3/... n_rules=1 patterns=[("/youtube/*", 1)] injection_count=0 auth_len=0
```

After the fix it carries both, and the Bearer applies:

```
path=/calendar/v3/... n_rules=3 patterns=[("/youtube/*",1),("/calendar/*",1),("/batch/calendar/*",1)] injection_count=1 auth_len=260
```

All five services (Gmail dedicated + legacy, Calendar, Drive, Tasks, YouTube) return 200 after the fix; selective-mode agents still correctly get `access_restricted`.

**Local checks.** `cargo fmt --check`, `cargo build`, `cargo test`, and `cargo clippy -- -D warnings` all pass for `apps/gateway`.

**On test coverage.** I didn't add a dedicated unit test with this change: `resolve_rules` (MITM) and the HTTP-proxy resolve path have no existing unit harness, and `tests/integration.rs` requires a seeded Postgres (`#[ignore]`d). A focused unit test is doable by seeding the `connect:` and `app_injection:` caches (so token decrypt/refresh is skipped) and asserting `resolve_rules` merges the secret and OAuth rules — but it needs a couple of internal structs made `pub(crate)`, which I didn't want to do unprompted. Happy to add that unit test, or an `#[ignore]`d integration test, in whichever form matches your conventions — just let me know which you'd prefer.

**AI Disclosure.** Code drafted with Claude Code, reviewed and modified by me. Root cause was found by instrumenting the gateway and reading the resolution path; I verified the repro, the fix, and the before/after behavior against a local build myself before submitting.

### DIFF
--- a/apps/gateway/src/connect.rs
+++ b/apps/gateway/src/connect.rs
@@ -636,17 +636,35 @@ impl PolicyEngine {
         let candidates = narrow_connections_by_path(app_connections, hostname, request_path);
         let app_connections: &[db::AppConnectionRow] = &candidates;
 
-        // Single connection — use it directly
+        // Single connection — use it directly. Its rules always merge (they
+        // self-select by path at apply time), but the winner metadata is
+        // dropped when the provider does not serve this request's path — a
+        // lone Calendar connection on a `/youtube/` request must not donate
+        // its granular policy, finalizer, or host rewrite.
         if app_connections.len() == 1 {
-            return self
-                .resolve_connection_injections(
-                    &app_connections[0],
-                    hostname,
-                    organization_id,
-                    project_id,
-                    cache,
-                )
-                .await;
+            let conn = &app_connections[0];
+            let mut result = self
+                .resolve_connection_injections(conn, hostname, organization_id, project_id, cache)
+                .await?;
+            if let AppConnectionResult::Rules {
+                provider,
+                rewrite_host,
+                connection_label,
+                finalizer,
+                body_transform,
+                session_policy,
+                ..
+            } = &mut result
+            {
+                if !provider_serves_request(provider, hostname, request_path) {
+                    *rewrite_host = None;
+                    *connection_label = None;
+                    *finalizer = None;
+                    *body_transform = None;
+                    *session_policy = None;
+                }
+            }
+            return Ok(result);
         }
 
         // Multiple connections — check for ambiguity per provider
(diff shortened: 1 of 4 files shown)


guyb1:
thanks @jbaruch ! I pushed a commit on top of yours that keeps your merge approach and completes it for a few edge cases.

will merge soon!

PR #430, opened 2026-07-16
feat(web): support a generic OIDC login provider via env vars

glifocat:

YES

Feature. Discussed in #181.

Self-hosted OAuth mode only works with Google (checked at v1.41.0,
`fe0981f0`). The provider is hardcoded in three places: the NextAuth config
only builds a `Google` provider
(`apps/web/src/lib/auth/nextauth-config.ts`), the sign-in call is
`nextAuthSignIn("google")` (`apps/web/src/lib/auth/auth-provider.tsx`), and
the startup check treats "no `GOOGLE_CLIENT_ID`" as misconfigured
(`apps/web/src/proxy.ts`). So a self-hoster with Keycloak, Authentik, Pocket
ID, or Dex cannot use their own identity provider, and fully offline
deployments cannot use multi-user mode at all (#181).

## What is the new behavior?

Four new env vars configure a generic OIDC provider next to Google:

- `OIDC_ISSUER`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`: standard OIDC
  discovery. The provider is active when all three are set.
- `OIDC_PROVIDER_NAME`: login button label, defaults to "SSO".

Behavior:

- Google is unchanged and stays first when both providers are configured, so
  existing deployments see no difference.
- The active provider id and label flow to the client through the existing
  runtime-config mechanism (entrypoint-written `runtime-config.json` in
  Docker, env fallback in local dev). The login button label follows
  `OIDC_PROVIDER_NAME`; Google keeps its branded button.
- The `oauth-misconfigured` setup error now fires only when
  `NEXTAUTH_SECRET` is set and neither Google nor OIDC is configured.
- Callback URL for IdP registration: `<APP_URL>/api/auth/callback/oidc`.
- The provider sets `checks: ["pkce", "state", "nonce"]` explicitly. Without
  it, Auth.js 5.0.0-beta.30 builds the authorize URL with only
  `code_challenge`, and spec-strict IdPs reject the callback with
  `invalid_state` (I hit this with Pocket ID before adding it). Lenient IdPs
  accept either; strict ones need this.
- The `/setup-error` page copy now describes both provider options instead
  of only the Google vars, matching the updated check.
- New vars documented in `.env.example` and the README, and declared in
  `turbo.json` so `turbo/no-undeclared-env-vars` stays clean.

One known limitation, kept deliberately to keep the diff small: the login
page still renders a single button. When both Google and OIDC are configured
the button starts the Google flow; the OIDC flow is still reachable at
`/api/auth/signin/oidc`. If you would rather have one button per configured
provider I am happy to extend the login page in this PR or a follow-up.

## How I verified it

- `pnpm check-types` (apps/web) passes on the branch.
- `eslint --max-warnings 0` passes on all changed files.
- Live end-to-end test against Pocket ID 2.10.0 (`pnpm dev`, OIDC-only
  config, no Google vars set): the login page renders "Continue with Pocket
  ID", the full flow completes (authorize, passkey login, callback, session),
  sign-out and re-login work, and the session sync upserts the user with the
  OIDC `sub` as `external_auth_id` and the IdP email (checked in Postgres).
  With `NEXTAUTH_SECRET` set and no provider configured, the setup-error
  page shows as expected.
- Not tested: a non-localhost plain-http issuer (the test issuer was
  `http://localhost`; https issuers are the normal case) and the
  both-providers-configured combination beyond unit-level reasoning.

## Additional context

- Cloud edition is untouched: its runtime config pins the existing Google
  behavior, and the Cognito path does not go through this code.
- Reverse-proxy deployments need `NEXTAUTH_URL` set to the external origin.
  The Docker image already ships `AUTH_TRUST_HOST=true`, so no code change
  was needed for proxy support; `.env.example` notes it.
- `docker/entrypoint.sh` interpolates `OIDC_PROVIDER_NAME` into
  `runtime-config.json` the same way the existing fields are written, so a
  label containing a double quote would break the JSON. That matches the
  current pattern for the other fields; I can add escaping if you want it.

Assisted by Claude; reviewed and verified by a human before submission.


mrrobertkent:
Ran this branch against Authentik on a live self-hosted instance behind TLS (merged onto current `main` — no conflicts, 18 commits behind). It works: the login button renders with `OIDC_PROVIDER_NAME`, and the sign-in redirect is correct.

```
302 -> https://<authentik>/application/o/authorize/
   client_id             = <client id>
   redirect_uri          = https://<instance>/api/auth/callback/oidc
   scope                 = openid profile email
   code_challenge_method = S256
```

`runtime-config.json` carried the new fields through as expected:

```json
{"authMode":"oauth","oauthConfigured":true,"authProviderId":"oidc","authProviderName":"Authentik"}
```

Two notes from the integration.

### `OIDC_ISSUER` must keep the provider's trailing slash

Worth a line in `.env.example`, because the failure mode doesn't point at the config. openid-client compares the `issuer` claim in the discovery document against the configured value verbatim. Authentik publishes:

```json
{"issuer": "https://<authentik>/application/o/<app-slug>/"}
```

Configure it without that trailing slash — which looks tidier, and matches how the value is usually written elsewhere — and sign-in dies at the callback with:

```
[auth][error] nP: "response" body "issuer" property does not match the expected value
```

With the slash, discovery is requested at `.../<app-slug>//.well-known/openid-configuration`. That doubled slash answers `301`, which is followed transparently, so it is harmless — the URL looks wrong but works, while the tidy-looking form fails. Something like "use the issuer exactly as your provider publishes it, including any trailing slash" would save people the round trip.

### On TLS, this needs a gateway-side fix too

Not a problem with this PR, but they interact. On an https deployment the browser's calls to the gateway (vault pair/status, approvals, cache invalidation) return `401`, because `validate_oauth` matches the session cookie by exact name `authjs.session-token` while Auth.js sends `__Secure-authjs.session-token` whenever the resolved `AUTH_URL`/`NEXTAUTH_URL` is https — which it must be, since providers require an https redirect URI.

That is invisible today because self-hosters overwhelmingly run local mode, where the cookie is never inspected. This PR is what changes that: it brings non-Google IdPs, and therefore more TLS multi-user deployments, into the path. I opened #475 for the gateway side; with both applied, Authentik login and the vault panels work together.

apps/web/src/lib/auth/auth-provider.tsx
```diff
@@ -25,14 +25,22 @@ const LocalAuthProvider = ({ children }: { children: ReactNode }) => {
       user: LOCAL_USER,
       signIn: async () => {},
       signOut: async () => {},
+      authProviderId: "",
```
ashroaban003, line 28: can we use interface here instead of passing 2 new variable ? authProvider: AuthProviderInfo | null; similarly for other places

apps/web/src/lib/runtime-config.ts
```diff
@@ -31,9 +45,24 @@ export const getRuntimeConfig = (): RuntimeConfig => {
...
+    const oidcConfigured = !!(
+      OIDC_ISSUER &&
+      OIDC_CLIENT_ID &&
+      OIDC_CLIENT_SECRET
+    );
     cached = {
       authMode: NEXTAUTH_SECRET ? "oauth" : "local",
-      oauthConfigured: !!GOOGLE_CLIENT_ID,
+      oauthConfigured: !!GOOGLE_CLIENT_ID || oidcConfigured,
+      authProviderId: GOOGLE_CLIENT_ID
```
ashroaban003, line 56: similarly , is it possible to mutualize to using autproviderinfo instead 2 variable??

PR #431, opened 2026-07-16
fix(docker): reap the gateway — tini as PID 1 plus a supervising entrypoint

tenequm:
> **Maintainer note.** Thanks @tenequm — this was a good catch and the diagnosis
> holds up. The defunct `node` process you observed is real and reproducible on a
> stock build, the reasoning about PID 1 not reaping is correct, and the
> nodejs/docker-node *Handling Kernel Signals* guidance you cited is the right
> reference. Your `tini` change is kept exactly as you wrote it.
>
> While verifying it I found the fix needs a second half to actually clear the
> zombie you reported, so I pushed a commit on top rather than asking you to redo
> it. Nothing of yours was reverted or rewritten. Details below.

## The problem

`docker/entrypoint.sh` starts the gateway in the background, sets a `TERM`/`INT`
trap to shut it down, and then `exec`s the web server. That `exec` does two
damaging things:

1. **Node becomes PID 1** and inherits the gateway as a direct child. Node reaps
   only what it spawned itself through `child_process`, so when the gateway exits
   it stays `<defunct>` forever — the process reported in this PR.
2. **The trap is discarded.** `exec` replaces the shell, taking its signal
   handlers with it. The graceful gateway shutdown written three lines earlier
   has never run — not once.

## Why `tini` alone is not enough

This is the part that is invisible from the outside, so it is worth spelling out.

`tini` reaps processes **reparented to PID 1** — genuine orphans. Our gateway is
never one: `entrypoint.sh` backgrounds it and *then* `exec`s node, so node
inherits it as a direct child and remains its parent for the container's whole
life. The reaping responsibility stays with node, which does not do it.

Measured on the real image, with this PR's `tini` change in place:

```
PID   PPID  STAT COMMAND
  7      1  S    next-server        <- node, child of tini
 15      7  Z    onecli-gateway     <- still a zombie, now at PPID 7
```

`tini` moved the zombie one level down the tree; it did not remove it. Adding an
init is correct and worth doing — it is what handles true orphans and restores
signal delivery — but the entrypoint has to stop `exec`ing for the reported
symptom to go away.

## What was added

`docker/entrypoint.sh` now supervises both services instead of exec'ing one of
them. Everything above the service start is untouched (migrations, key
generation, runtime config).

- Both services run as background children of a shell that stays alive to
  supervise them, reap them, and honor its own trap. `tini` sits above that.
- The `TERM`/`INT` trap now actually fires, so `docker stop` runs the graceful
  shutdown path the script always described.
- Waiting is done by polling liveness rather than `wait -n`. `wait -n` looks like
  the right tool and is not: it only reports children that exit *after* it is
  called, so a service dying during startup — a bad key, an unreachable database
  — is missed entirely. Measured with one child already dead, `wait -n` blocked
  for 27 s until the *other* child exited. Polling asks the question that
  matters, and `sleep` stays interruptible so the trap fires immediately.
- The exit code propagates the way `exec node` used to.

Written against busybox `ash` specifically, since that is what `/bin/sh` is on
Alpine. Three behaviors were tested there rather than assumed: `wait -n`'s return
value is not the child's status (it gave `129` where bash gives `7`), `wait $PID`
stays accurate even after the child was reaped, and `set -e` aborts on a non-zero
`wait` before a following `RC=$?` can run.

## One behavior change

**A dying gateway now stops the container** (exit 1). Today it dies unnoticed and
the container keeps serving with no proxy until the healthcheck eventually
notices. Under `restart: unless-stopped` the container will now restart instead
of limping. This is the intended improvement, but it is a change in behavior and
worth knowing about before merging.

## Verification

Built and driven on the real all-in-one image — a process-model change earns a
real container rather than a synthetic probe. A/B against the same image built
from `main`:

| | before | after |
| --- | --- | --- |
| PID 1 | `next-server` | `tini` |
| tree | node → gateway | `tini` → `entrypoint.sh` → {gateway, next-server} |
| zombies at startup | 1 (`node <defunct>`) | 0 |
| kill the gateway | 2nd zombie, container keeps running | exits 1, reason logged |
| `docker stop` | trap never runs | trap runs, returns immediately |
| healthcheck | healthy | healthy |

Also `sh -n docker/entrypoint.sh` and `docker compose config`.

## For a reviewer

- `main` was merged into this branch first — it was six commits behind. None of
  those six touched `docker/`, so nothing was at risk of being reverted, but the
  merge means the verification below ran against current `main` rather than a
  four-week-old tree.
- Nothing from the original commit was removed. The `apk add tini` +
  `ENTRYPOINT` lines and their comment are as authored.
- The busybox notes above are the load-bearing ones; if any of them is wrong on
  another Alpine version, the failure is a container that hangs rather than one
  that exits, so it is worth a second pair of eyes.


guyb1:
great! thank you @tenequm 
merged

PR #434, opened 2026-07-19
fix(gateway): send client_id on OpenAI OAuth token refresh

### LINKED ISSUE #433: Gateway can never refresh vaulted ChatGPT (Codex) OAuth tokens: refresh request omits client_id
## Summary

The gateway's automatic OpenAI OAuth refresh (`refresh_openai_oauth_if_expired` in `apps/gateway/src/secret_inject.rs`) always fails, because the refresh request to `https://auth.openai.com/oauth/token` sends only `grant_type` and `refresh_token`. The endpoint requires `client_id` and rejects the request with 400:

```
WARN onecli_gateway::secret_inject: openai oauth token refresh failed, using expired token error=openai oauth token refresh failed (400 Bad Request): {
  "error": {
    "message": "Missing 'client_id'",
    "type": "invalid_request_error",
    "param": "client_id",
    "code": "missing_required_parameter"
  }
```

The gateway then falls back to injecting the expired token, so every request to `chatgpt.com` gets 401:

```
WARN ... onecli_gateway::gateway::websocket: WebSocket: upstream rejected upgrade host=chatgpt.com:443 status=401
```

Net effect: a vaulted ChatGPT-subscription session (e.g. for driving the Codex CLI through the gateway) hard-expires roughly every 10 days and requires a full interactive re-login, despite the docs saying "The gateway automatically refreshes the access token when it expires using the stored refresh token — no manual re-authentication needed."

Downstream symptom worth noting for searchability: Codex CLI reacts to the 401 by attempting to rewrite its own `~/.codex/auth.json`, which is the gateway's read-only sentinel stub, so end users see `Reconnecting... 2/5: Read-only file system (os error 30)`.

## Reproduce

1. Vault a ChatGPT-subscription `auth.json` (`onecli secrets create --file ... --host-pattern chatgpt.com`), use it through the gateway.
2. Wait for the access token to expire (~10 days).
3. Observe the 400/401 pairs above in gateway logs; all Codex traffic fails until manual re-auth.

## Fix

Include the Codex CLI's public OAuth `client_id` (`app_EMoamEEZ73f0CkXaXp7hrann`, hardcoded in the open-source Codex CLI, e.g. `codex-rs/login/src/auth/manager.rs`) in the refresh form — it identifies the app that issued the vaulted session and is not a secret. One-line change; PR follows.

Verified against `main` as of 2026-07-19 (`apps/gateway/src/secret_inject.rs`, `refresh_openai_oauth_token`).

morganfamilyagents:
Fixes #433.

`auth.openai.com/oauth/token` rejects a `refresh_token` grant that omits `client_id` with `400 Missing 'client_id'`, so `refresh_openai_oauth_token` has never once succeeded. The failure is quiet by design — the caller logs a warning and falls through to the *expired* token — so a vaulted ChatGPT session works until its access token expires (~10 days), then every `chatgpt.com` request 401s until someone deletes the secret and re-authenticates by hand.

---

> **Maintainer note.** @morganfamilyagents diagnosed this precisely — the endpoint, the missing field, the client id, and the log evidence in #433 were all correct, and the fix is unchanged in substance. I pushed a follow-up commit adding the tests and the endpoint verification, since the PR description noted a local Rust build wasn't available. Thanks for the unusually complete report; the reproduction steps and the downstream `Read-only file system (os error 30)` symptom made this quick to confirm.

---

## Verified against the live endpoint

Probed with a bogus refresh token (public endpoint, no auth):

| Request | Response |
| --- | --- |
| form, no `client_id` — **before this PR** | `Missing 'client_id'` / `missing_required_parameter` |
| form + `client_id` | `token_expired` — parsed, moved past `client_id` |
| **JSON** + `client_id` | `token_expired` — identical |
| form + **wrong** `client_id` | `Invalid client specified.` / `invalid_client` |

Row 1 reproduces the reported error exactly. Row 4 shows the id is genuinely validated and that this one is right.

Row 3 is worth recording: the Codex CLI sends this grant as **JSON**, not form-encoded (`codex-rs/login/src/auth/manager.rs`), so the form body looked like it might be rejected regardless of the added field. The endpoint accepts both identically — so the existing encoding stays and the change remains one field.

## Why a constant

The vaulted `auth.json` carries no `client_id` — confirmed in Codex's own `TokenData` — so it cannot be read from the credential. The value identifies OpenAI's first-party public CLI and is hardcoded in the open-source client. That is unlike the OAuth providers in `apps.rs`, which take a `client_id_env` because an *operator* supplies their own app there.

## Tests

The form construction is split into a small pure function, because the request itself needs the network; the tests then assert the field set and pin the client id as a literal.

Both were mutation-checked: renaming the key to `clientId` fails one, mistyping a single character of the id fails the other. That matters here because either mistake otherwise surfaces only as an OpenAI error inside a warning, days later.

## Not changed

`id_token` and `last_refresh` are still not persisted from the refresh response — deliberately. `default_interceptions.rs` intercepts Codex's *local* self-refresh with a synthetic `200` so Codex stamps its own `last_refresh`, and notes that `id_token` is intentionally not returned. Injection reads only `access_token` and `account_id`. The upstream refresh was the broken half.

The fall-through-to-expired-token behavior also stays: once the refresh works, that path only fires for a genuinely dead refresh token, where re-authenticating *is* the right answer.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets`, and `cargo test` (483 + 8) all green with **no features** — the standalone OSS build.

### DIFF
--- a/apps/gateway/src/secret_inject.rs
+++ b/apps/gateway/src/secret_inject.rs
@@ -236,6 +236,12 @@ pub(crate) async fn refresh_openai_oauth_if_expired(
     }
 }
 
+/// Public OAuth client id of the Codex CLI (the app that issued the vaulted
+/// ChatGPT session). auth.openai.com rejects refresh_token grants that omit
+/// `client_id` with 400 "Missing 'client_id'", so it must accompany every
+/// refresh. Not a secret — it is hardcoded in the open-source Codex CLI.
+const OPENAI_CODEX_CLIENT_ID: &str = "app_EMoamEEZ73f0CkXaXp7hrann";
+
 /// Refresh an OpenAI OAuth access_token using the refresh_token.
 async fn refresh_openai_oauth_token(
     refresh_token: &str,
@@ -244,6 +250,7 @@ async fn refresh_openai_oauth_token(
         .post("https://auth.openai.com/oauth/token")
         .timeout(std::time::Duration::from_secs(10))
         .form(&[
+            ("client_id", OPENAI_CODEX_CLIENT_ID),
             ("grant_type", "refresh_token"),
             ("refresh_token", refresh_token),
         ])
--- a/apps/gateway/src/secret_inject.rs
+++ b/apps/gateway/src/secret_inject.rs
@@ -236,24 +236,40 @@ pub(crate) async fn refresh_openai_oauth_if_expired(
     }
 }
 
-/// Public OAuth client id of the Codex CLI (the app that issued the vaulted
-/// ChatGPT session). auth.openai.com rejects refresh_token grants that omit
-/// `client_id` with 400 "Missing 'client_id'", so it must accompany every
-/// refresh. Not a secret — it is hardcoded in the open-source Codex CLI.
+const OPENAI_TOKEN_URL: &str = "https://auth.openai.com/oauth/token";
+
+/// Public OAuth client id of the Codex CLI — the app that issued the vaulted
+/// ChatGPT session we are refreshing.
+///
+/// `auth.openai.com` rejects a `refresh_token` grant that omits it with
+/// 400 `Missing 'client_id'`, so without this the refresh can never succeed and
+/// the session hard-expires with the access token (~10 days).
+///
+/// A constant rather than configuration: it identifies OpenAI's own first-party
+/// CLI, is hardcoded in the open-source Codex client, and the vaulted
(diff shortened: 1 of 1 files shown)


guyb1:
@morganfamilyagents thank you for pointing this one!
merged

PR #445, opened 2026-07-24
feat: add self-managed Airbyte connection

### LINKED ISSUE #444: Add a first-class self-managed Airbyte connection
Hi OneCLI team — first, this is a great app. I’ve been using it as the credential boundary for agent-driven API work, and Airbyte exposed a useful gap that seems worth supporting directly.

I’d like to contribute a first-class **self-managed Airbyte connection**. This is specifically for a customer-operated Airbyte instance, not Airbyte Cloud.

The proposed connection would:

- accept the self-managed instance origin, an Airbyte application client ID, and client secret in one connection;
- exchange those client credentials for Airbyte’s short-lived bearer token and refresh it automatically;
- inject the bearer token only for the exact configured HTTPS authority (including a non-default port) and `/api/public/v1` path;
- keep separate Airbyte instances as separate connections, identified by normalized Public API URL;
- avoid persisting per-instance client credentials as a provider-wide app configuration;
- use a 10-second token-request timeout and never follow redirects;
- permit private/internal instance origins for self-managed installations, while requiring admin/owner authority when OneCLI RBAC is active;
- start with project-scoped connections and origin-root deployments only (reverse-proxy subpaths would be a follow-up).

I have an implementation and focused security tests ready, including exact host/port isolation, path-boundary enforcement, dynamic app-availability policy enforcement, token error redaction, and the alternate Airbyte `grant-type` / `grant_type` request forms. The full repository test, check, and production build suites are passing locally.

I’d be happy to adjust the shape if you’d prefer a different convention for dynamic self-managed hosts or connection identity.

thethomasjfellows:

YES

Feature and documentation.

OneCLI does not have a first-class way to connect a self-managed Airbyte instance. A generic bearer secret can work only until Airbyte’s short-lived access token expires, while generic header injection cannot perform Airbyte’s JSON-body client-credentials exchange.

Closes #444.

## What is the new behavior?

This adds a first-class **self-managed Airbyte connection**. It is intentionally not an Airbyte Cloud connection.

A project admin enters one instance origin, application client ID, and client secret. OneCLI exchanges and refreshes the short-lived token, then injects it only on the exact configured HTTPS authority (including a non-default port) under `/api/public/v1`.

The implementation also:

- keeps each Airbyte instance as a separate endpoint-identified connection;
- prevents per-instance client credentials from becoming provider-wide app config;
- enforces dynamic Airbyte hosts in app-availability policy checks across CONNECT, plain HTTP proxy, and WebSocket paths;
- requires admin/owner authority when RBAC is active and starts with project-scoped connections;
- uses a 10-second token timeout, does not follow redirects, and redacts credentials from errors;
- rejects reverse-proxy subpaths for now and fails closed outside the supported Public API path;
- documents setup and the explicit self-managed-versus-Cloud boundary in the UI, gateway guidance, and project docs.

## Additional context

OneCLI is a great app, and I built this after using it as the credential boundary for an Airbyte workflow. The integration was reviewed specifically for host/port isolation, private-network SSRF boundaries, token refresh behavior, and policy enforcement.

Local QA is green:

- `pnpm check`
- `pnpm test` — 585 API tests, 518 gateway unit tests, and 8 gateway integration tests
- `pnpm build` — optimized web build and release gateway build

I kept the first version deliberately narrow: self-managed Airbyte, project scope, HTTPS origins, and no reverse-proxy subpath support. I’m happy to adjust the dynamic-host shape if there is a preferred internal convention.


thethomasjfellows:
Quick live-QA follow-up: after deploying the fork build against an existing self-managed Airbyte connection, I found one more recovery edge case. A stored Airbyte token can be rejected before its recorded expiry (for example after a restart or revocation), so the gateway now remints an Airbyte application token on a cold/cache-miss resolution and then caches it normally. The focused Airbyte tests and the full lint/type/format gate are green after that change.

That kind of credential-boundary behavior is exactly why I like OneCLI—it's a great app, and it gives this integration one reliable place to recover instead of making every agent rediscover the token flow.

guyb1:
hey @thethomasjfellows , i just commented in the issue :)

thethomasjfellows:
Thanks! I’m running a local Codex agent through OneCLI. It uses Airbyte’s Public API to inspect connections and sync status, and when authorized, trigger or retry syncs.

The main need is for OneCLI to handle the client-credentials exchange and token refresh so the agent never sees the client secret or bearer token. The current narrow self-managed connection design covers the use case well.

guyb1:
great! @thethomasjfellows sounds awesome
will review it really soon and will update!

apps/gateway/src/apps.rs
```diff
@@ -1462,6 +1492,9 @@ pub(crate) fn rewrite_host(
 /// Returns true if the provider has any host rule that injects an Authorization header.
 /// Providers using only credential_headers (e.g., Datadog) return false.
 pub(crate) fn needs_access_token(provider: &str) -> bool {
+    if provider == "airbyte" {
```
ashroaban003, line 1495: its better to find alternative where we mention this inside all_providers(). so it follows regular iteration path like below, stacking if(x) return true is not good,  the correct behavior is to derive this from the all_provider definitions

apps/gateway/src/connect.rs
```diff
@@ -900,12 +910,19 @@ impl PolicyEngine {
             })
             .ok();
 
+        // Dynamic self-managed Airbyte hosts are stored per connection. Reject
+        // any mismatch before token resolution, rule building, or caching.
+        if conn.provider == "airbyte" && !airbyte_connection_matches_host(conn, authority) {
```
ashroaban003, line 915: this is too specific to airbyte , is it possible to create something like **connection_host_field** , connection_host_mismatch which throws error dynamically . which manage things like airbyte . Like below how they manage for credential_host_field, this way we remove specific if ()conditions below just for airbyte

PR #461, opened 2026-07-29
fix anthropic oauth token help

### LINKED ISSUE #459: Anthropic secret form doesn't say which OAuth token to use
Connections → LLMs → Anthropic accepts any `sk-ant-oat…` value, but two different Anthropic OAuth tokens share that prefix:

- the short-lived token Claude Code rotates internally — expires every few hours
- the long-lived token from `claude setup-token` — the one intended for external tools

Nothing in the UI distinguishes them, so pasting the keychain token produces 401s a few hours later with no indication why. This is exactly what #322 reported. That issue was resolved with a promise to document it:

> on the docs: fully on us. the current flow makes it way too easy to pull the wrong token from the keychain since both live there. we'll add a clear note pointing to `claude setup-token` so the next person isn't in this loop.

`grep -rn "setup-token"` across the repo returns no hits, so that note never shipped. Happy to send a PR adding it to the secret dialog.

erichanwang:
fixes #459

## changes
- points Anthropic OAuth users to `claude setup-token` for a long-lived token

## verification
- `pnpm --filter @onecli/db prisma generate`
- `pnpm --filter @onecli/web check-types`
- `pnpm --filter @onecli/web lint`
- `pnpm exec prettier --check 'apps/web/src/app/(dashboard)/connections/_components/secret-dialog.tsx'
- pre-push `pnpm check` hook passed

### DIFF
--- a/apps/web/src/app/(dashboard)/connections/_components/secret-dialog.tsx
+++ b/apps/web/src/app/(dashboard)/connections/_components/secret-dialog.tsx
@@ -887,7 +887,13 @@ export const SecretDialog = ({
                       ) : (
                         <p className="text-muted-foreground text-xs">
                           {type === "anthropic" ? (
-                            "Paste your API key or OAuth token from the Anthropic Console."
+                            <>
+                              Paste your API key, or run{" "}
+                              <code className="text-[11px]">
+                                claude setup-token
+                              </code>{" "}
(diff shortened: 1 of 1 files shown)


erichanwang:
closing this draft as duplicate of #460, which already covers the same setup-token guidance.

PR #474, opened 2026-07-31
fix(gateway): require API key auth even in AUTH_MODE=local

### LINKED ISSUE #263: Local gateway API does not enforce ONECLI_API_KEY when bound to TCP loopback
## Summary

The OneCLI gateway's HTTP API on `127.0.0.1:10254` returns sensitive data with **no authentication enforcement** — `Authorization: Bearer <key>` is accepted but completely optional. `ONECLI_API_KEY` exists in the auth model and the CLI dutifully attaches it, but the gateway returns the same 200 responses with no header, a fake Bearer, or the real one.

In a single-user macOS install this means: any process running as the local user (browsers, dev tools, future malware, accidentally-trusted scripts) can read every connected app's metadata, every secret name + host pattern + scope, every agent + secret-mode binding, etc. via plain HTTP `GET`. The "authenticated as admin@localhost" indicator the gateway returns is process-internal narrative; it doesn't gate any external HTTP request.

## Repro (current latest, gateway image `1.22.0`, CLI `1.6.0`)

```
$ curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:10254/api/secrets
200

$ curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer wrongkey" http://127.0.0.1:10254/api/secrets
200

$ curl -s http://127.0.0.1:10254/api/secrets | head -c 200
[{"id":"7c6cde2f-…","name":"Anthropic","type":"anthropic","hostPattern":"api.anthropic.com",…
```

`/api/agents` and `/api/apps` behave the same — wide open. (`/api/auth/status` returns 400 without a session cookie, but that's about session shape, not access control to the data endpoints.)

## Why this matters

OneCLI's positioning is "containers never receive raw API keys; the vault is the only path to credentials" — that's the value proposition for forks like NanoClaw using OneCLI as a credential gateway for agent containers. But the threat model that posture *implies* (compromise of an agent container ≠ credential disclosure) is undercut by the local API being readable by every other process on the host. A user who installs OneCLI to keep their Anthropic/Google OAuth tokens away from agent containers is probably also not expecting a `localhost:10254` GET to leak the secret list to any browser tab they have open.

The macOS Keychain comparison is *almost* right — Keychain also trusts processes running as the user — but Keychain mediates per-item with system-prompt fallbacks for sensitive items. OneCLI on TCP loopback has no such fallback; it's pure trust-anyone-on-the-loopback.

## Proposed fixes (any one solves it)

1. **Enforce `ONECLI_API_KEY` on all API endpoints when bound to TCP.** The CLI already attaches it; the web UI can attach it via cookie/session set during the initial auto-login. Single-user installs continue to "just work" because the key lands in `~/.onecli/credentials/api-key` and the CLI/UI both have access. The change is only externally-visible: third-party HTTP requests without the key get 401. Estimated impact: small, no UX change for legitimate users.

2. **Offer a `bind: unix-socket` transport** (e.g. `~/.onecli/gateway.sock` with `0600` perms). UID-level access control delegated to the kernel; matches the security posture macOS users implicitly expect from a "local-only" service. Containers needing access bind-mount the socket. Compose file gains a `volumes:` line; the CLI's `api-host` config gains a `unix:///path` form.

3. **At minimum: a `--bind localhost-trusted` mode that requires explicit opt-out**, with the default mode rejecting unauthenticated requests. Users who genuinely want the current behaviour (e.g. local-only single-user installs that prefer convenience) opt in.

## Scope of impact

- Single-user macOS installs (dominant use case): primary risk vector is malicious or buggy local processes, including unprivileged user-space tools that haven't been audited for security.
- Multi-user Linux hosts: any logged-in user can read every other user's gateway. The 1.20.0 "configurable app/gateway ports for multi-user hosts" feature implicitly acknowledges multi-user setups exist; auth enforcement matters more there.
- The cloud-mode work in 1.21.0 (cloud-only apps framework, user provisioning schema, credential header injection) appears to anticipate auth-aware deployment, but the local mode hasn't been tightened to match.

## Related

- [`#244` (1.20.0)](https://github.com/onecli/onecli/issues/244) added `SecretInput` and the OpenAI secret type — UI work that exposes more sensitive values in the same un-gated endpoint surface.
- The CLI already issues an API key during install. Enforcing it would be additive, not a redesign.

Happy to test fixes if you ship a candidate.

erichanwang:
fixes #263

## changes

This is a two-sided fix — a gateway-side hole, and a web-side dependency on that hole that had to be closed first.

**Gateway (`apps/gateway/src/auth.rs`)**: `AuthUser::from_request_parts` tries `validate_api_key` (checks `Authorization: Bearer oc_...` against the DB) and only falls back to `validate_request` → `validate_local` when no such key is presented. `validate_local` then unconditionally authenticated the caller as `local-admin` — full owner — with **no credential of any kind**. Every gateway route (vault pair/status, cache invalidate, pending approvals, approval decisions) goes through this same extractor, so in `AUTH_MODE=local` (the shipped single-user-dev default) any bare request reaching the gateway's loopback port was authenticated as the admin. `validate_local` is removed; local mode now rejects a request that already failed the API-key check rather than auto-trusting it — that's the only credential path in local mode (there's no login/session mechanism to fall back to).

**Web (`apps/web/src/lib/gateway-auth.ts`)**: the web app's own gateway calls relied on exactly this bypass — `getGatewayFetchOptions` sent `headers: {}`. Closing the gateway hole without fixing this would have broken vault pairing, cache invalidation, and approvals for every user. `getGatewayFetchOptions` now fetches/provisions the caller's real API key via the existing `ensureApiKey`-backed server action (the same one the API Keys settings page uses) and attaches it as `Authorization: Bearer oc_...`. This works unchanged under `oauth` mode too — sending a valid key alongside the session cookie just gives the gateway a stronger credential to check first.

**Trust model, before → after:**
- Before: `AUTH_MODE=local` + no `Authorization` header → authenticated as `local-admin` (full owner), unconditionally.
- After: `AUTH_MODE=local` + no valid `Authorization: Bearer oc_...` → `401 Unauthorized`. The web app now always sends a valid key. A human hitting the gateway directly (curl, no web UI) gets their key from the existing API Keys settings page (`ensureApiKey`) — no functionality removed, just no longer free for the taking.

Updated two stale comments in `apps/web/src/lib/api/{cache,approvals}.ts` that described the old "edition-aware, no need to borrow an API key" design, which no longer matches.

## verification
- `cargo fmt --manifest-path apps/gateway/Cargo.toml --check`
- `cargo test --manifest-path apps/gateway/Cargo.toml auth::` — 4 passed, including new regression test `validate_request_local_mode_rejects_bare_request` (bare request in local mode → `Err`)
- `pnpm --filter web check-types`
- `pnpm --filter web lint`
- `git diff --check`
- `pnpm check` (full monorepo lint + check-types + format:check, also ran again via the pre-commit hook on commit)

### DIFF
--- a/apps/gateway/src/auth.rs
+++ b/apps/gateway/src/auth.rs
@@ -1,7 +1,9 @@
 //! Gateway authentication for browser requests.
 //!
 //! Supports two modes controlled by the `AUTH_MODE` env var:
-//! - `local`: bypasses JWT validation, looks up the "local-admin" user directly.
+//! - `local`: no session mechanism at all (single-user dev, no login) — the
+//!   only accepted credential is an `Authorization: Bearer oc_...` API key,
+//!   checked by `validate_api_key` before mode-specific logic even runs.
 //! - `oauth` (default): validates a NextAuth session cookie JWT (HS256).
 
 use std::sync::OnceLock;
@@ -142,26 +144,17 @@ async fn validate_api_key(pool: &PgPool, headers: &HeaderMap) -> Option<AuthUser
 /// The caller resolves the project ID from the user's membership.
 async fn validate_request(pool: &PgPool, headers: &HeaderMap) -> Result<String, AuthError> {
     match auth_mode() {
-        "local" => validate_local(pool).await,
+        // Local mode has no cookie/session mechanism to fall back to (no
+        // login flow exists). A request reaching here already failed the API
+        // key check above, so it is unauthenticated — reject it rather than
+        // auto-trusting it as local-admin. This is the only credential path
+        // in local mode, so it must hold even when the gateway is bound to
+        // loopback: anything else on the same host could otherwise reach it.
+        "local" => Err(AuthError(
+            "missing API key (Authorization: Bearer oc_...)".to_string(),
+        )),
         _ => validate_oauth(pool, headers).await,
     }
-}
-
-// ── Local mode ───────────────────────────────────────────────────────────
-
-async fn validate_local(pool: &PgPool) -> Result<String, AuthError> {
-    let user = db::find_user_by_external_auth_id(pool, "local-admin")
-        .await
-        .map_err(|e| {
-            warn!(error = %e, "local auth: db error");
-            AuthError("internal error".to_string())
-        })?
-        .ok_or_else(|| {
-            warn!("local auth: local-admin user not found");
-            AuthError("user not found".to_string())
-        })?;
-
-    Ok(user.id)
 }
 
 // ── OAuth mode ───────────────────────────────────────────────────────────
@@ -258,5 +251,22 @@ mod tests {
     #[test]
     fn parse_cookie_empty() {
         assert_eq!(parse_cookie("", "authjs.session-token"), None);
+    }
+
+    /// Regression test for the loopback auth bypass (local gateway API did
+    /// not enforce ONECLI_API_KEY when bound to TCP loopback): a bare
+    /// request with no `Authorization` header must be rejected in `local`
+    /// mode instead of silently authenticating as local-admin.
+    #[tokio::test]
+    async fn validate_request_local_mode_rejects_bare_request() {
+        std::env::set_var("AUTH_MODE", "local");
+        let pool = sqlx::postgres::PgPoolOptions::new()
+            .connect_lazy("postgres://localhost/nonexistent")
+            .expect("lazy pool never fails to construct");
+        let headers = HeaderMap::new();
+
+        let result = validate_request(&pool, &headers).await;
+
+        assert!(result.is_err());
     }
 }
--- a/apps/web/src/lib/api/approvals.ts
+++ b/apps/web/src/lib/api/approvals.ts
@@ -1,8 +1,8 @@
 // Pending-approval value picker — list and resolve held requests through the
 // gateway. Like the 1Password client, these hit the gateway directly (different
 // base URL + auth) rather than the typed JSON API, so they use
-// getGatewayApiUrl() + getGatewayFetchOptions() (edition-aware: Cognito Bearer +
-// X-Project-Id in cloud, cookie credentials in OSS).
+// getGatewayApiUrl() + getGatewayFetchOptions(), which authenticates as the
+// acting user via their own API key (see gateway-auth.ts).
 import { getGatewayApiUrl } from "@/hooks/use-vault-status";
 import { getGatewayFetchOptions } from "@/lib/gateway-auth";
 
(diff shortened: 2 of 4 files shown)


mrrobertkent:
Running this branch in production on a self-hosted TLS deployment, and it fixes a second problem beyond the one in the description — worth recording, because that problem makes multi-user mode unusable on its own.

## The web half also repairs browser→gateway auth in `oauth` mode

`validate_oauth` cannot authenticate an Auth.js session in any release. It decodes the cookie with `jsonwebtoken` under `Algorithm::HS256`, i.e. a three-segment JWS, but Auth.js issues an **encrypted** token — the built bundle contains `EncryptJWT`/`jwtDecrypt` with `A256CBC-HS512` and `dir`, and `nextauth-config.ts` sets no custom `jwt.encode`/`decode`. With a real session the gateway gets:

```
WARN onecli_gateway::auth: oauth auth: JWT decode failed error=Base64 error: Invalid symbol 46, offset 174
```

Symbol 46 is `.` — a five-segment JWE being read as a three-segment JWS.

`git log -L '/^async fn validate_oauth/,/^\/\/ ── Helpers/' -- apps/gateway/src/auth.rs` returns exactly one commit, and that same commit already pinned `next-auth@5.0.0-beta.30` and shipped `gateway-auth.ts` with empty headers. So the session path appears never to have worked; it is masked on default installs because `validate_local` short-circuits before it is reached.

**This PR's web half sidesteps that entirely.** Once the browser sends `Authorization: Bearer oc_…`, `validate_api_key` matches before the mode branch and the token format never comes up. Measured on the same instance, before and after:

```
before   /v1/vault/onepassword/status   401   (oauth auth: JWT decode failed)
after    /v1/vault/onepassword/status   200   {"connected":false,…}
```

1Password pairing, vault status, and the approvals poll all work now; before, the vault panels were permanently 401 while the rest of the dashboard was fine.

## Deployment it was verified on

Community edition, `AUTH_MODE=oauth`, behind Traefik with TLS, Authentik as the IdP via the generic OIDC provider from #430. The gateway half was exercised too: a bare request to the gateway returns 401, and the same request with a valid `oc_…` key returns 200.

## Two notes for reviewers

**The exposure argument is weaker than it looks.** My first reaction was that attaching the key moves a long-lived credential into page JavaScript. It does not introduce that: `overview/_components/api-key-card.tsx` already fetches the raw `oc_` key into React state and offers copy-to-clipboard, so it is in page JS today regardless.

**The gateway half is a genuine behaviour break** for anyone driving the local gateway with bare `curl` — which is #263's own repro. Intentional and argued in the description, but it may be worth a release note.

I opened #475 for the cookie-name half of the session path (`__Secure-` prefix on TLS). With this PR applied that path is unreachable, so #475 is optional — happy for it to be closed if you would rather not carry code for a path that no longer runs.

erichanwang:
Thanks for the production verification. This confirms the API-key browser-to-gateway path fixes the OAuth-mode session failure described here. I’ll keep the Auth.js JWE/cookie-prefix follow-up separate in #475 rather than expanding this PR’s scope.

PR #483, opened 2026-08-07
fix(gateway): match Proxy-Authorization scheme case-insensitively

tenequm:

YES

Bug fix (gateway).

`extract_agent_token` matches the `Proxy-Authorization` auth-scheme byte-for-byte:

```rust
let encoded = value.strip_prefix("Basic ")?.trim();
```

RFC 7235 section 2.1 defines `auth-scheme` as a case-insensitive token, and PySocks emits the lowercase form:

```
Proxy-Authorization: basic <base64>
```

PySocks is what httplib2 uses for proxy support, so this affects every httplib2-based client. That includes `google-api-python-client`, and therefore any Python MCP server or script reaching a Google API through the gateway.

The failure mode is silent rather than loud. `extract_agent_token` returns `None`, the CONNECT carries no agent identity, and `intercept` stays `false` (`apps/gateway/src/gateway.rs:786-812`), so the gateway plain-tunnels the connection. No 407, no error: the request leaves the sandbox **unauthenticated, with no credential injected**, and the only trace is `project_id="-"` in the gateway logs. A client that should have been credentialed instead reaches the upstream API as an anonymous caller.

An *invalid* token is handled correctly (407). Only the casing mismatch takes this path, because a dropped token is indistinguishable from "no proxy auth was offered at all".

Reproduce with any httplib2 caller through the gateway, or directly:

```
Proxy-Authorization: basic <base64("x:aoc_...")>   ->  token dropped, tunnel unauthenticated
Proxy-Authorization: Basic <base64("x:aoc_...")>   ->  token extracted, credentials injected
```

Prior art in this repo: `validate_api_key` (`apps/gateway/src/auth.rs:119-120`) already handles the same class of problem for `Authorization`, via `strip_prefix("Bearer ").or_else(|| strip_prefix("bearer "))`. That form covers two casings but still misses `BEARER`, which is why this PR compares the scheme rather than enumerating spellings.

## What is the new behavior?

The scheme is compared case-insensitively. Everything else is unchanged: `Bearer` in any casing still returns `None`, and the base64 payload is still trimmed and decoded exactly as before. I walked the input shapes (`Basic`/`basic`/mixed case, no-space, trailing-space, leading-space, double-space, tab-separated, wrong scheme) and the only behavioral difference is the intended one.

Four tests added alongside the existing `extract_agent_token` cases: lowercase `basic`, mixed-case `BaSiC`, and lowercase `bearer` still rejected. The mixed-case test is deliberate: it fails against the `or_else(strip_prefix("basic "))` shape, so a future revert to that idiom will not pass silently.

## Checks

Run locally on this branch:

- `cargo test` in `apps/gateway`: 521 passed, 0 failed
- `cargo fmt --check` and `cargo clippy --all-targets`: clean
- `pnpm build`: 2/2 tasks successful
- `pnpm check`: 9/9 tasks successful


tenequm:

tenequm:
recheck

PR #487, opened 2026-08-09
fix(gateway): inject Anthropic OAuth tokens correctly to resolve credential_not_found

### LINKED ISSUE #482: Agent-grants (v1.44.0+) never inject: correctly published policy_rules_v2 grant still 401s credential_not_found
## Problem

A `type: anthropic` secret (Claude subscription OAuth token) granted to an agent via the dashboard's agent-grants UI (LLMs tab, toggle on) is never actually injected. Every call 401s with `credential_not_found`, even though the grant is correctly saved and shown as attached.

## Steps to Reproduce

1. `onecli secrets create --name Anthropic --type anthropic --value sk-ant-oat01-... --host-pattern api.anthropic.com`
2. Dashboard → Agents → Default Agent → LLMs tab → toggle the secret on (confirmed via network tab: `PUT /v1/agents/<id>/grants/secrets/<secretId>` → 200 OK)
3. `onecli run -- claude -p "hi"`

## Expected Behaviour

Claude Code authenticates using the granted credential and responds normally.

## Observed Behaviour

```
onecli: gateway connected. Starting claude...
Failed to authenticate. API Error: 401 No credentials configured for api.anthropic.com in OneCLI.
```

Also reproduces with a bare `curl` through the gateway (no `onecli run`, no Claude Code), using the agent's own access token:

```bash
curl -sS -x http://<agent-access-token>@localhost:10255 \
  --cacert /tmp/onecli-ca.pem \
  -X POST https://api.anthropic.com/v1/messages \
  -H "x-api-key: placeholder" -H "anthropic-version: 2023-06-01" -H "content-type: application/json" \
  -d '{"model":"claude-sonnet-5","max_tokens":10,"messages":[{"role":"user","content":"hi"}]}'
# -> {"error":"credential_not_found", ...} HTTP 401
```

Reproduces identically for the Default Agent and a second, custom-created agent — not specific to one agent identity.

I checked the underlying data directly in Postgres (secret + agent share the same `project_id`, the `policy_rules_v2` grant rule is `published`/`enabled` at the current max generation for the project, with matching `policy_rule_identities`/`policy_rule_targets` rows) — the grant looks correctly configured, but injection still never happens.

## Environment

- Self-hosted (`curl -fsSL onecli.sh/install | sh`), single-project local install
- Gateway: pulled `ghcr.io/onecli/onecli:1.36.0`, but `/api/health` reports `"version":"1.45.0"`
- `onecli` CLI 2.2.5
- Credential: Claude Pro/Max subscription token via `claude setup-token`

Adityakk9031:
Fixes #482

**Description**
Resolves an issue where Anthropic OAuth tokens (`sk-ant-oat...`) granted via the agent-grants UI failed to inject, resulting in a `401 credential_not_found` error from the Gateway.

**Root Cause**
The injection logic for Anthropic OAuth tokens mistakenly used `Injection::ReplaceHeader` to inject the `authorization` header. `ReplaceHeader` only applies if the header is already present in the incoming request. Since Claude Code or a bare `curl` typically sends `x-api-key: placeholder` and omits the `authorization` header, the injection was skipped (`injection_count == 0`). This caused the Gateway to forward the request unauthenticated and return `credential_not_found` when it inevitably received a 401 from Anthropic.

**Changes**
- Switched to `Injection::SetHeader` for the `authorization` token to ensure it is always injected.
- Added `Injection::RemoveHeader` for `x-api-key` to strip the placeholder from the request before forwarding it upstream.
- Updated the `build_injections_anthropic_oauth` unit test in `secret_inject.rs` to assert the new injection behavior.

### DIFF
--- a/apps/gateway/src/secret_inject.rs
+++ b/apps/gateway/src/secret_inject.rs
@@ -23,10 +23,15 @@ pub(crate) fn build_injections(
         "anthropic" => {
             let is_oauth = decrypted_value.starts_with("sk-ant-oat");
             if is_oauth {
-                vec![Injection::ReplaceHeader {
-                    name: "authorization".to_string(),
-                    value: format!("Bearer {decrypted_value}"),
-                }]
+                vec![
+                    Injection::SetHeader {
+                        name: "authorization".to_string(),
+                        value: format!("Bearer {decrypted_value}"),
+                    },
+                    Injection::RemoveHeader {
+                        name: "x-api-key".to_string(),
+                    },
+                ]
             } else {
                 vec![
                     Injection::SetHeader {
@@ -364,12 +369,18 @@ mod tests {
     #[test]
     fn build_injections_anthropic_oauth() {
         let injections = build_injections("anthropic", "sk-ant-oat-test-token", None, None);
-        assert_eq!(injections.len(), 1);
+        assert_eq!(injections.len(), 2);
         assert_eq!(
             injections[0],
-            Injection::ReplaceHeader {
+            Injection::SetHeader {
                 name: "authorization".to_string(),
                 value: "Bearer sk-ant-oat-test-token".to_string(),
+            }
+        );
+        assert_eq!(
+            injections[1],
+            Injection::RemoveHeader {
+                name: "x-api-key".to_string(),
             }
         );
     }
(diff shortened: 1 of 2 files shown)


Adityakk9031:
@guyb1 have a look

PR #495, opened 2026-08-16
Fix/485 port qualified host matching

### LINKED ISSUE #485: Generic secret hostPattern accepts ports but gateway strips them before matching
## Summary

Generic secrets can be created with a port-qualified `hostPattern`, such as
`example.internal:3000`, but the gateway strips the request port before resolving
secrets. The stored pattern therefore never matches.

This prevents distinct credentials from being assigned to services hosted on
different ports of the same hostname—a common pattern for local and self-hosted MCP
servers.

## Reproduction

1. Create a generic secret with:

   - `hostPattern`: `example.internal:3000`
   - `pathPattern`: `/api/mcp`
   - injection: `Authorization: Bearer {value}`

2. Attach the secret to an agent.
3. Send an HTTP request through the authenticated OneCLI proxy:

   ```text
   POST http://example.internal:3000/api/mcp
   ```

4. The gateway forwards the request without injecting the secret.

Observed gateway output:

```text
HTTP_PROXY scheme=http injection_count=0 policy_count=0
credential not found ... status=401
```

The agent was confirmed to have access to the secret.

## Cause

Both proxy paths remove the port before credential resolution:

- CONNECT handler:
  https://github.com/onecli/onecli/blob/84ccaf74ce6350f8925675457b48272c29f32c1a/apps/gateway/src/gateway.rs#L777-L793
- Plain HTTP handler:
  https://github.com/onecli/onecli/blob/84ccaf74ce6350f8925675457b48272c29f32c1a/apps/gateway/src/gateway.rs#L947-L965

`connect::resolve` then caches and resolves exclusively by hostname:

https://github.com/onecli/onecli/blob/84ccaf74ce6350f8925675457b48272c29f32c1a/apps/gateway/src/connect.rs#L1540-L1600

The matcher performs an exact or wildcard comparison against that stripped hostname:

https://github.com/onecli/onecli/blob/84ccaf74ce6350f8925675457b48272c29f32c1a/apps/gateway/src/connect.rs#L1745-L1772

The API validation rejects URLs and paths but does not reject an exact
`hostname:port` pattern, so the configuration is accepted but cannot work at runtime.

## Why removing the port is not always safe

A hostname-only workaround is unsafe when separate services on different ports use
different credentials.

For example:

```text
example.internal:3000 → credential A
example.internal:3200 → credential B
```

If both secrets are changed to `example.internal` and have overlapping paths, both
can match the same request and attempt to set `Authorization`. The selected value
becomes order-dependent and could send a credential to the wrong service.

## Proposed semantics

Would the project accept port-qualified host patterns with these semantics?

- `example.internal` continues matching the hostname on any port.
- `example.internal:3000` matches only port 3000.
- Wildcard behavior remains backward compatible.
- Credential-resolution cache keys include the normalized effective port so rules
  cannot leak across ports.
- HTTP proxy and CONNECT/MITM paths use the same endpoint matching.
- Secret policy targets use the same endpoint semantics as injection.
- Validation accepts supported `hostname:port` forms and rejects malformed ones.

Provider detection and TLS certificate generation should continue using the bare
hostname.

## Versions

- Reproduced on OneCLI server `1.38.0`
- Current upstream `main` at `84ccaf7` / release `1.45.0` still contains the same
  hostname-only resolution path

AromalBiju1:
## I have read the CONTRIBUTING.md file.
YES

Bug fix.

Fixes #485.

Generic secrets can be created with a port-qualified `hostPattern` (e.g.
`example.internal:3000`), but the gateway strips the port from the request
authority before resolving credentials. `host_matches` only ever sees the
bare hostname, so a port-qualified pattern can never match — the secret is
silently unusable and requests get `credential not found ... status=401`.

## What is the new behavior?
- The port-inclusive authority is now passed into `connect::resolve` at both
  the CONNECT and plain-HTTP proxy paths, instead of the port-stripped
  hostname.
- `host_matches` splits both the request host and the pattern into
  host/port. A pattern with a port (`example.internal:3000`) now requires
  an exact port match. A bare-hostname pattern (`example.internal`) still
  matches any port, so existing behavior is unchanged.

Verified locally by creating a generic secret with `hostPattern:
example.internal:3000`, granting it to an agent, and sending a request
through the proxy:

- Before: `injection_count=0` — secret never matched.
- After: `injection_count=1` — secret injected correctly.

Added two unit tests covering both cases. Full `connect::` suite passes
(35/35), `cargo build` and `pnpm check` pass clean.

## Additional context
Didn't change the secret-creation validation to reject malformed
`hostname:port` patterns (mentioned as a "nice to have" in the issue) —
happy to add that in a follow-up if wanted.


chiptoe-svg:
Thanks for taking this on — the approach matches what we found when filing #485, and passing the port-inclusive authority at *both* `resolve` sites covers our real-world case (plain-HTTP MCP servers on one host across several ports).

A few notes from a close read:

1. The hunk that adds the new tests also drops the `#[test]` attribute from the existing `host_exact_match` — it silently stops running.
2. Worth confirming `mitm.rs`'s in-tunnel re-resolution also sees the ported authority, or CONNECT-path port-qualified secrets would match at CONNECT but not inject.
3. Two documentable edges: `host:80` won't match `http://host/…` (implicit port ≠ explicit pattern), and an unbracketed IPv6 pattern like `::1` mis-splits in `split_host_port`.

Happy to test the branch against our live setup — seven port-qualified plain-HTTP servers behind one hostname, which is the scenario from the issue.

AromalBiju1:
Thanks for the detailed review - all three are spot-on!

Fixed the dropped #[test] on host_exact_match. Got caught by the cargo fmt diff, sorry!
And yeah, the problem you noted regarding mitm.rs - the re-resolution in the tunnel still stripped the port prior to resolving from cache, so it would match a CONNECT path secret but never re-inject it per each request. Fixed it with the same pattern, preserving the port till the resolution stage.
Good edge cases - host:80 vs implicit-port URLs, and also the unbracketed IPv6 patterns. Going to leave those out of scope for now and just document as such, but let me know if you'd prefer me to solve them here.

Going to push an updated commit soon. Your real-world testing on that seven-ports-one-hostname case would be very helpful!

AromalBiju1:

chiptoe-svg:
Reporter of #485 here. The fix looks right to me, and I wanted to put the downstream numbers on the record in case they're useful for review.

The part you caught on your own is the one that would have bitten us hardest: the MITM tunnel re-resolving with the port stripped. A secret that matches on the CONNECT path but then fails to re-inject per request is exactly the failure that looks fine in a smoke test and goes intermittent in production.

Concretely what this unblocks downstream — a multi-user NanoClaw install that fronts all agent egress through the gateway. We run **seven services on one hostname, separated only by port**: `host.docker.internal:3000, 3200, 8011, 8012, 8765, 8766, 8767`, across three agent groups. Because a port-qualified `hostPattern` can't match today, we distinguish credentials with a *fake* path discriminator — a bare hostname plus `--path-pattern /something/*` — and then have to verify each server genuinely ignores paths before trusting it. Port-qualified matching replaces that workaround with the obvious key.

I haven't run your branch: we're pinned to a released gateway in production and I'm not swapping the binary under live users. Happy to confirm the behavior on our topology once it's in a release.

*Written with Claude's help; reviewed and edited before posting.*

AromalBiju1:
Thanks for putting the real numbers on record — the fake-path-discriminator workaround makes a lot more sense now knowing what it's actually standing in for. No rush testing on your end; whenever it lands, would be good to hear how it holds up on your topology.

PR #524, opened 2026-08-27
fix(cli-auth): dynamically resolve apiHost for self-hosted instances

### LINKED ISSUE #496: Get Started page tells the CLI to use the gateway port, so `onecli auth login` fails
## Steps to reproduce

1. Start OneCLI locally (`curl -fsSL https://onecli.sh/install | sh` or `docker compose`)
2. Open the dashboard and go to Install ("Get Started").
3. Copy the commands from step 2, "Install & sign in". They read:

   curl -fsSL onecli.sh/cli/install | sh
   onecli config set api-host http://localhost:10255
   onecli auth login --api-key oc_...

4. Run them.

**Result:** login fails with an error that says nothing about the cause.

```json
{
  "error": "could not verify API key: getting user: Bad Request",
  "code": "ERROR"
}
```

**Expected**: the copied commands work as-is.

## Solution

I managed to auth the CLI by specifying 10254 port instead of 10255:

```
onecli config set api-host http://localhost:10255
```

Apparently, the UI displays the wrong port.

## Environment

- OneCLI 1.45.0

### Why it happens 

  `API_URL` defaults to `http://localhost:10255`, the gateway port, in three places: `apps/web/src/lib/env.ts:27`,
  `packages/api/src/lib/env.ts:18`, and `apps/web/next.config.js:189` (when `API_DOMAIN` is unset). `.env.example:56` records the
  assumption behind it: `# OSS: localhost:10255 (gateway serves API too)`.

  `getInstallInfo()` passes it through as `apiUrl` (`apps/web/src/lib/actions/secrets.ts:80`) into the `config set api-host` line
  (`install-content.tsx:86`). The CLI verifies the key with `GET /user`, which on 10255 lands in the Axum `fallback()`
  (`apps/gateway/src/gateway.rs:698`) and comes back as a bodyless 400, so the CLI has nothing to report but "Bad Request".

  Two notes:

  - The published image is built without `API_DOMAIN`, so `localhost:10255` is baked into the bundle and operators can't override it
  at runtime. `apps/web/src/lib/env.ts` also doesn't read plain `API_URL`, unlike `packages/api/src/lib/env.ts`.
  - The `hint` in the working response resolves through `app-origin.ts` and is correct; `apiUrl` doesn't, so it's also wrong for any
  self-host reached over a LAN IP, proxy, or tunnel, not just for the port.

  ### Suggested fix

  Resolve `apiUrl` and `appUrl` in `getInstallInfo` as `configuredAppUrl() ?? originFromHeaders(await headers()) ?? APP_URL`, using
  `API_URL` only under `IS_CLOUD`.

  Then correct the `API_URL` defaults in both `env.ts` files and `next.config.js:189`, plus the `.env.example:56` comment.
  `GATEWAY_API_URL` / `GATEWAY_API_DOMAIN` stay as they are.

  Worth considering separately: giving `fallback()` a body that names the app port, since the gateway is the only component positioned
  to spot this misconfiguration.

Adityakk9031:
Fixes #496. Dynamically resolves apiHost/appUrl in getInstallInfo from the browser's request headers on self-hosted instances when no external URL is configured.

### DIFF
--- a/apps/web/src/lib/actions/secrets.ts
+++ b/apps/web/src/lib/actions/secrets.ts
@@ -4,6 +4,8 @@ import { db } from "@onecli/db";
 import { resolveWorkspaceContext } from "@/lib/actions/resolve-user";
 import type { ResolveOptions } from "@/lib/actions/resolve-user";
 import { apiOrigin, appOrigin } from "@onecli/api/lib/public-origins";
+import { originFromHeaders } from "@onecli/api/lib/app-origin";
+import { headers } from "next/headers";
 import {
   listSecrets,
   createSecret as createSecretService,
@@ -76,10 +78,13 @@ export const getInstallInfo = async (options?: ResolveOptions) => {
     });
   }
 
+  const headerList = await headers();
+  const requestOrigin = originFromHeaders(headerList);
+
   return {
     apiKey: keyResult.apiKey,
-    appUrl: appOrigin(),
-    apiUrl: apiOrigin(),
+    appUrl: appOrigin(requestOrigin),
+    apiUrl: apiOrigin(requestOrigin),
   };
 };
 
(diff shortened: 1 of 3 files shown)


Adityakk9031:

Adityakk9031:
@aminsamir45 have a look

PR #525, opened 2026-08-27
fix(gateway): limit openai host expansion to oauth secrets only

### LINKED ISSUE #490: OpenAI API-key injection breaks fresh Codex OAuth login with invalid_client
## Summary

When an `openai` secret in API-key mode is stored for `api.openai.com`, OneCLI 1.45.0 also applies it to `auth.openai.com` and injects it as an `Authorization: Bearer ...` header.

That added header causes a fresh Codex ChatGPT OAuth authorization-code exchange at `https://auth.openai.com/oauth/token` to fail with:

```text
401 Unauthorized
invalid_client: Invalid client specified.
```

This prevents `codex login` from completing whenever Codex traffic is routed through the authenticated OneCLI gateway. It also sends the API key to an OpenAI-owned hostname outside the secret's configured `hostPattern`.

## Environment

- OneCLI server: `1.45.0`
- Current `main`: `84ccaf74ce6350f8925675457b48272c29f32c1a`
- Codex CLI: `0.147.0`
- OneCLI secret: type `openai`, API-key mode, `hostPattern=api.openai.com`
- Client: WSL Ubuntu, with HTTPS routed through the authenticated OneCLI proxy

## Steps to reproduce

1. Store an OpenAI API-key secret in OneCLI with type `openai` and `hostPattern=api.openai.com`.
2. Grant it to the agent used by an authenticated OneCLI proxy connection.
3. Route Codex HTTPS traffic through that proxy.
4. Start from logged-out Codex state and run `codex login`.
5. Complete the browser authorization flow.

The browser returns to Codex, but the token exchange fails with `token_exchange_failed` / `invalid_client`.

## Controlled evidence

I sent the same synthetic, already-expired authorization-code form to the token endpoint in three conditions; no live authorization code or credential was used in this control:

| Request path | Result |
| --- | --- |
| Direct to `auth.openai.com` | `401 token_expired` |
| Through the OneCLI gateway | `401 invalid_client` |
| Direct, with a synthetic `Authorization: Bearer not-a-real-key` header | The same `401 invalid_client` response as the gateway path |

The OneCLI gateway logged the proxied token request with `injections_applied=1`.

The direct request proves that the public Codex client ID and form shape reach the intended token-validation path. Adding only the Authorization header reproduces the gateway failure exactly.

## Cause

`secret_host_patterns` expands every `openai` secret to `*.openai.com`, regardless of whether it contains an API key or a ChatGPT OAuth credential:

https://github.com/onecli/onecli/blob/84ccaf74ce6350f8925675457b48272c29f32c1a/apps/gateway/src/secret_inject.rs#L162-L187

API-key mode then injects the secret as an Authorization bearer value:

https://github.com/onecli/onecli/blob/84ccaf74ce6350f8925675457b48272c29f32c1a/apps/gateway/src/secret_inject.rs#L43-L86

This also contradicts the default-interception contract, which says real Codex login requests are forwarded to `auth.openai.com` untouched:

https://github.com/onecli/onecli/blob/84ccaf74ce6350f8925675457b48272c29f32c1a/apps/gateway/src/default_interceptions.rs#L84-L89

The broad OpenAI host expansion was introduced in commit `1633c21` / PR #438.

## Expected behavior

- An API-key-mode secret configured for `api.openai.com` should not be injected into `auth.openai.com` OAuth requests or ChatGPT subscription endpoints.
- A real Codex authorization-code exchange should be forwarded without an API-key Authorization header.
- ChatGPT OAuth-mode secrets can retain their separate ChatGPT host coverage.
- The existing synthetic handling for the `onecli-managed` refresh placeholder should continue to work.

One possible fix is to make extra OpenAI host patterns depend on `metadata.authMode`, rather than only `secret_type`, and to ensure that a real `/oauth/token` request which is not synthetically intercepted bypasses secret injection.

## Workaround

Bypass OneCLI for the subscription authentication and ChatGPT hosts while keeping `api.openai.com` proxied:

```text
NO_PROXY=auth.openai.com,chatgpt.com,.chatgpt.com
```

## Related issues

- #433 concerns refreshing a vaulted ChatGPT OAuth credential and is not this failure.
- #425 concerns Codex model-backend routing after OneCLI-managed login and is not this failure.

Adityakk9031:
Fixes #490. Limits the OpenAI host pattern expansion to secrets that specify oauth in metadata.authMode. This prevents OpenAI API-key-mode secrets from being injected into OAuth authentication requests to auth.openai.com.


Adityakk9031:
@guyb1 have a look

guyb1:
great catch! thanks @Adityakk9031 
merged

PR #534, opened 2026-08-31
feat: Microsoft 365 integration (Outlook mail + calendar via Graph)

deweysasser:
## Summary

Ships a working **Microsoft 365** integration (Outlook mail + calendar via Microsoft Graph), replacing the locked "Outlook Mail"/"Outlook Calendar" placeholder cards with one connectable app card.

Design decision (spec in `docs/superpowers/specs/2026-08-30-microsoft-365-integration-design.md`): **one provider (`microsoft-365`), one card** — Graph serves mail and calendar from one host under `/v1.0/me/...`, so per-capability granularity lives in the permissions/policy layer rather than a provider split.

- **OAuth helper** (`packages/api/src/apps/oauth/microsoft.ts`): identity-platform v2.0, `common` tenant, `offline_access` refresh tokens, `/v1.0/me` metadata — mirrors the Google helper idiom
- **App definition + registry**: `microsoft-365` card (scopes: User.Read, Mail.ReadWrite, Mail.Send, Calendars.ReadWrite), BYOC config with `MICROSOFT_CLIENT_ID`/`MICROSOFT_CLIENT_SECRET` env defaults; outlook placeholders removed
- **Permissions catalogue**: 22 tools (9 read / 13 write) over Graph mail + calendar endpoints, `/v1.0/users/*` alias patterns
- **Gateway**: `microsoft-365` provider (exact-host Bearer injection on `graph.microsoft.com`) + `MICROSOFT_REFRESH` (handles Microsoft's refresh-token rotation via the existing shared path)
- **First JS test harness in the repo**: vitest in `packages/api` (14 tests), now also run in CI

## Security fix included

Review of this branch surfaced a pre-existing gateway matcher issue: `/prefix/*` path patterns also matched the **bare prefix**, so an allow rule for `get_message` (`/v1.0/me/messages/*`) would have implicitly granted `GET /v1.0/me/messages` (full listing). Fixed in `2ede30a`:

- `/prefix/*` now requires a non-empty trailing segment (bare and trailing-slash collection paths no longer match)
- `"/*"` is special-cased as match-all (preserves AWS query-protocol tools and existing DB rules)
- docker/vercel/confluence/lambda catalogue tools that relied on bare-prefix matching got explicit bare-path `aliasPatterns`, preserving their grant coverage

Note: user-authored custom policy rules of the form `/x/*` narrow under the new semantics — allow rules tighten (fail closed); a block rule targeting a bare collection path would need the bare path added explicitly.

## Not in this PR (manual prerequisites)

- Azure App Registration (common tenant, delegated Mail/Calendar scopes, app-connect callback URI) and `MICROSOFT_CLIENT_ID`/`MICROSOFT_CLIENT_SECRET` in deployment secrets
- Manual E2E: connect → Graph call through gateway → refresh rotation → policy block on `sendMail` (checklist in `docs/superpowers/plans/2026-08-30-microsoft-365-integration.md` Task 6)

## Test plan

- [x] vitest 14/14 (`pnpm --filter @onecli/api test`) — OAuth exchange incl. edge cases (missing refresh_token/expires_in/scope), registry, permissions catalogue
- [x] cargo 356 unit + 7 integration; clippy `-D warnings`; fmt clean
- [x] Full lint + check-types + format across JS packages
- [x] New CI job runs the vitest suite on PRs
- [ ] Manual E2E after Azure registration (above)

https://claude.ai/code/session_01FyVqyZQV8EzTVa4RLtgHtc


chiptoe-svg:
Thanks for this. The shape mirrors the Google provider closely and the permissions catalogue makes the read/write split easy to gate in policy. A few notes from a university deployment (Microsoft Entra work tenant with Conditional Access), plus one structural suggestion.

1. Tenant is hard-coded to common. The authorize URL, token URL, and the gateway’s MICROSOFT_REFRESH.token_url all use login.microsoftonline.com/common. Single-tenant app registrations, which are the default for IT-reviewed apps at universities and enterprises, are rejected on that endpoint (AADSTS50194). A tenant field in the BYOC config, defaulting to common, would cover both cases. The refresh side needs the same value, and since RefreshConfig is a static table, that probably means a per-connection override for the token URL, or a documented limitation until the framework supports it.

2. Mail.Send is in the default scopes. Every connection consents to sending on the user’s behalf even when the deployment’s policy blocks the sendMail tool. The permissions catalogue already distinguishes read from write, so making Mail.Send opt-in at connect time would match that model and reduce what a tenant admin has to approve.

3. Refresh-time Conditional Access. On managed tenants a refresh can come back with interaction_required or a claims challenge rather than a token. The spec’s “expired token forwarded, upstream 401 reaches the agent” is a reasonable first behavior, but it would help to log the error field from the token endpoint so the operator can tell “re-consent needed” from “token expired”.

4. Consider splitting the PR. It carries three separable pieces: the provider itself, the gateway prefix-matcher security fix (2ede30a), and the repo’s first vitest harness plus its CI job. The matcher fix in particular deserves to land on its own and quickly. Split out, this PR would be roughly a thousand lines including the docs.

Offer: the E2E checklist in the plan is against a personal Microsoft account, which does not exercise Conditional Access, refresh-token rotation under a managed tenant, or single-tenant registration. I can run it against a real Entra work tenant with those policies once this is close to mergeable, and report back here.

PR #535, opened 2026-08-31
feat(gateway): env-gated Anthropic upstream override for org LLM proxies

GuitaristForEver:
## Why

Organizations that front the Anthropic API with their own proxy (e.g. Azure APIM in front of Azure AI Foundry, which speaks the native Anthropic Messages API) need the gateway to deliver Anthropic-bound traffic to that proxy instead of `api.anthropic.com` — while still making every policy, approval, and secret-injection decision against the real host.

## What

Env-gated override, **off by default** (both variables empty ⇒ zero behavior change):

- `ANTHROPIC_UPSTREAM_URL` — base URL of the proxy (host + optional path prefix, e.g. `https://proxy.example.com/anthropic`). When set, requests targeting `api.anthropic.com` are re-targeted to the proxy **after** policy evaluation and header injection, so the injected `x-api-key` (the stored secret — the proxy's key in this setup) rides along unchanged.
- `ANTHROPIC_UPSTREAM_EXTRA_HEADERS` — comma-separated `name=value` pairs added to every overridden request (e.g. APIM attribution headers). Also applied when a harness points straight at the proxy via `ANTHROPIC_BASE_URL` — the URL is already right there, so only the headers are added.

Both are plumbed through docker-compose with empty defaults.

## Testing

- 8 unit tests in `anthropic_upstream.rs` (URL parsing, host matching, header parsing, direct-path headers)
- `cargo test` green
- Running in production-like deployment routing live traffic through Azure APIM → AI Foundry

### DIFF
--- a/apps/gateway/src/anthropic_upstream.rs
+++ b/apps/gateway/src/anthropic_upstream.rs
@@ -1,0 +1,236 @@
+//! Optional, env-gated override of the Anthropic upstream.
+//!
+//! Some self-host installs must reach Claude through an organization proxy
+//! (e.g. Azure API Management fronting Azure AI Foundry) that speaks the
+//! NATIVE Anthropic Messages API on a different host, often under a path
+//! prefix, and requires extra attribution headers. The harness inside a
+//! sandbox hardcodes `api.anthropic.com`, and the gateway already MITMs that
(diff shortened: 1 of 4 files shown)


GuitaristForEver:

PR #536, opened 2026-08-31
feat(runner): operator-level extra env vars for sandbox containers

GuitaristForEver:
## Why

Operators sometimes need to hand every sandboxed agent an environment value the control plane does not know about. The driving use case: pointing every harness at an organization LLM proxy with `ANTHROPIC_BASE_URL=https://…` (pairs with the gateway-side override in #535, but useful independently).

## What

`RUNNER_SANDBOX_EXTRA_ENV` — comma-separated `KEY=VALUE` pairs injected into every sandbox container the runner creates.

Precedence, deliberate:
1. queue payload env
2. **operator extras (this feature)** — operator wins over payload on conflict
3. control-channel variables (`SANDBOX_ID`, `RUNNER_WS_URL`, `SANDBOX_WS_TOKEN`) — always win, stay untouchable

Entries without `=` are dropped rather than injected as empty-named variables the daemon would reject.

## Testing

- Behavioral test: extras are injected, payload vars survive, control-channel vars untouched
- `pnpm check-types` and full runner suite green (234 tests)

### DIFF
--- a/apps/hosted-e2e/src/runner.ts
+++ b/apps/hosted-e2e/src/runner.ts
@@ -76,6 +76,7 @@ export const startTestRunner = async (
     agentImage: opts.config.agentImage,
     sandboxNetwork: opts.ids.network,
     networkInternal: false,
+    sandboxExtraEnv: {},
     wsPort,
     advertisedHost: opts.config.hostGatewayHost,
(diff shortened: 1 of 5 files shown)


GuitaristForEver:

GuitaristForEver:
recheck

PR #537, opened 2026-08-31
feat(runner): opt-in escape hatch for hosts that cannot run hardened sandboxes

GuitaristForEver:
## Why

Some nested/restricted Docker hosts cannot exec the agent image's entrypoint under the default hardened profile — container creation fails with `exec ...: operation not permitted` under `no-new-privileges`. On such hosts the runner is currently unusable, with no supported way forward.

## What

`RUNNER_SANDBOX_UNSAFE_SECCOMP` (default **false**). When explicitly enabled, sandbox containers are created with `seccomp=unconfined` + `apparmor=unconfined` instead of `no-new-privileges`.

This is deliberately loud about its tradeoff: the default seccomp profile is the syscall gate that keeps rootless podman inside the sandbox off the shared kernel, and the config docs and code comments say so at every layer. I fully expect maintainers may prefer a different shape here (e.g. a custom seccomp profile instead of unconfined) — happy to rework; the pain point is real.

- Default path unchanged, still pinned by the existing "never weakens seccomp" test
- New test pins that unconfined applies **only** behind the explicit opt-in

## Testing

- `pnpm check-types` and full runner suite green (234 tests)
- Verified on a restricted Azure Docker host where the default profile cannot exec the entrypoint

### DIFF
--- a/apps/hosted-e2e/src/runner.ts
+++ b/apps/hosted-e2e/src/runner.ts
@@ -76,6 +76,7 @@ export const startTestRunner = async (
     agentImage: opts.config.agentImage,
     sandboxNetwork: opts.ids.network,
     networkInternal: false,
+    sandboxUnsafeSeccomp: false,
     wsPort,
     advertisedHost: opts.config.hostGatewayHost,
     maxSandboxes: opts.maxSandboxes ?? 4,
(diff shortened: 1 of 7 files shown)


GuitaristForEver:

GuitaristForEver:
recheck

PR #541, opened 2026-09-01
fix(gateway): bound upstream response-header wait and rotate suspect client pools

### LINKED ISSUE #493: Gateway upstream requests can remain pending indefinitely and require restart to recover
## Summary

A running OneCLI gateway entered a state where HTTPS requests to GitHub through the credential-injecting MITM path stopped completing. Some clients completed the client-facing TLS handshake with OneCLI but received no HTTP response; Git/libcurl-GnuTLS sometimes failed earlier with a non-proper TLS termination.

Restarting OneCLI cleared the runtime state. After recovery, the same requests completed normally without changing Git's HTTP version, TLS version, certificate verification, proxy configuration, credential grant, or target repository.

The initiating cause of the stalled upstream state is unknown. An unhealthy pooled HTTP/1.1 keep-alive connection or other stuck outbound state is plausible but not proven. The deterministic defect is that the gateway waits on upstream `reqwest` operations without a connection or response-header deadline and has no mechanism to rotate the shared client/pool after such a stall. A single bad upstream operation can therefore remain pending indefinitely and may require a process restart to recover.

## Impact

- Credential-injected HTTP/API traffic for an affected destination can stop completing.
- The client can receive a misleading secondary error such as a TLS handshake termination even when local CA trust and ordinary connectivity are valid.
- The gateway remains superficially healthy: its process and listener stay alive, and unrelated health/resource checks may remain normal.
- Recovery currently requires restarting the OneCLI container, interrupting all users of the gateway.

## Environment

- OneCLI image: `ghcr.io/onecli/onecli:1.45.0`
- Image digest: `sha256:d0177458b1f9ecece4abbe9abb6c5f925475357c1734f50a675d83a2ef9c8687`
- OneCLI source revision label: `84ccaf74ce6350f8925675457b48272c29f32c1a`
- Deployment: Docker, restart policy `unless-stopped`
- Host: Linux 6.8.0, x86_64
- Git: 2.43.0 (`1:2.43.0-1ubuntu7.3`)
- Git HTTP transport: libcurl-GnuTLS
  - `libcurl3t64-gnutls=8.5.0-2ubuntu10.11`
  - `libgnutls30t64=3.8.3-1.1ubuntu3.6`
- curl control client: curl/libcurl 8.5.0 with OpenSSL 3.0.13 and nghttp2 1.59.0
- Proxy listener: standard local OneCLI gateway port 10255
- Target observed in the field: public GitHub smart-HTTP endpoint for a public repository
- A GitHub credential connection was available to the agent and the MITM path applied one injection in a later sanitized cold-gateway control.

No credential value, proxy authorization value, agent/project identifier, request header, CA key, or private endpoint is included here.

## Chronology

The exact same OneCLI 1.45.0 image successfully served Hermes Agent update fetches on:

- 2026-08-03
- 2026-08-04
- 2026-08-05
- 2026-08-07
- 2026-08-09
- 2026-08-10
- 2026-08-14 (two successful update runs)

On 2026-08-16, three update attempts failed with:

```text
fatal: unable to access 'https://github.com/NousResearch/hermes-agent.git/':
gnutls_handshake() failed: The TLS connection was non-properly terminated.
```

This makes the problem intermittent. The image was not inherently unable to serve Git traffic after deployment.

## Field reproduction

The affected operation was an ordinary Git fetch through the configured OneCLI proxy:

```bash
timeout 60 git -C <public-repository-checkout> fetch origin main
```

Observed results during the incident:

- ordinary fetch hung until the external timeout or failed with the GnuTLS error above;
- CONNECT to the local OneCLI proxy returned `200`;
- forcing the Git client-facing leg to HTTP/1.1 was not a reliable fix: it succeeded in one early control but later HTTP/1.1 probes also stalled;
- disabling TLS verification was never attempted and is not needed;
- no repository-local or global Git HTTP/TLS override was present after recovery.

A neutral public reproduction target can use:

```bash
git ls-remote https://github.com/onecli/onecli.git HEAD
```

The original stuck runtime state was cleared by restart, so the exact field failure is no longer continuously reproducible.

## Protocol observations during the stuck state

A redacted probe recorded protocol milestones only; it did not retain or print request headers or credentials.

### Git/libcurl-GnuTLS

For both requested TLS 1.2 and TLS 1.3 client configurations:

- proxy CONNECT returned `200`;
- no certificate/negotiated-TLS milestone was observed;
- the operation remained pending for approximately 17 seconds until the probe's external timeout.

### curl/OpenSSL

For both TLS 1.2 and TLS 1.3, using HTTP/1.1 on the client-facing leg:

- proxy CONNECT returned `200`;
- ClientHello and ServerHello were observed;
- the OneCLI-issued certificate was received;
- the requested TLS version was negotiated;
- ALPN selected HTTP/1.1;
- no HTTP response arrived within 15 seconds.

The curl results place at least one failure after a successful client-facing TLS handshake. The GnuTLS error is therefore not sufficient evidence that CA trust, TLS 1.3, or the client-facing protocol was the primary cause.

An early local harness label called these curl cases `tls_handshake_timeout`; that label was corrected after reviewing the milestones. The issue intentionally reports the milestones rather than the stale label.

## Recovery observations

After OneCLI was restarted and startup had settled:

- curl TLS 1.2 + HTTP/1.1: success in about 0.28 seconds;
- curl TLS 1.3 + HTTP/1.1: success in about 0.28 seconds;
- Git through HTTP/1.1: success in about 0.9 seconds;
- Git with HTTP/2 on the client-facing Git-to-OneCLI leg: success in about 0.9 seconds;
- ordinary Git fetches worked without HTTP/TLS overrides.

A second controlled restart reached three consecutive successful checks for all of the following within approximately five seconds of container restart:

- OneCLI web health;
- neutral public HTTPS through the proxy;
- GitHub API through the proxy;
- GitHub smart-HTTP path through the proxy.

The restart removed pending tasks and rebuilt the shared `reqwest` connection pool. This supports an in-memory/runtime-state failure but does not uniquely identify the initiating fault.

## Other controls

- Basic public HTTPS through OneCLI worked after recovery.
- GitHub API, plain HTTP, and an existing internal credential-injection lane worked after recovery.
- A cold isolated gateway process performed the expected GitHub MITM path:
  - cache miss and database resolution;
  - one deferred connection;
  - successful client TLS handshake;
  - successful upstream TCP/TLS connection;
  - one injection applied;
  - HTTP response returned and the upstream connection was pooled.
- The container did not show CPU, PID, file-descriptor, or obvious memory exhaustion during diagnosis.
- Host Git, libcurl-GnuTLS, and GnuTLS packages were already at the latest versions available from the configured Ubuntu repositories.
- Direct no-proxy Git controls were inconclusive because this host may require the gateway for outbound access.

## Version/regression analysis

This does not appear to be a deterministic OneCLI 1.45.0 regression:

- the same image served successful Git updates through 2026-08-14;
- OneCLI v1.41.0 and v1.45.0 use the same relevant HTTP stack versions:
  - reqwest 0.12.28
  - hyper 1.8.1
  - rustls 0.23.37
  - Tokio 1.50.0
- the unbounded forwarding wait is present in both versions.

The client-facing Git leg did negotiate HTTP/2 in one recovered control, but Cargo feature inspection shows that OneCLI's upstream `reqwest` client is built without reqwest's `http2` feature. The upstream client pool is therefore HTTP/1.1. This rules out a pooled upstream HTTP/2 connection as the mechanism in this build. It does not identify the initiating cause; an unhealthy HTTP/1.1 keep-alive connection, a stuck request, or another transient outbound state remains plausible.

## Deterministic source-level reproduction

The gateway's normal forwarding path awaits the upstream request without a gateway-level bound:

```rust
let upstream_resp = upstream
    .send()
    .await
    .with_context(|| format!("forwarding to {url}"))?;
```

A focused test used a local upstream that:

1. accepted the TCP connection;
2. accepted the request;
3. never returned HTTP response headers.

Against the existing behavior, OneCLI did not terminate the operation itself. Only the test's outer 250 ms guard stopped it. A prototype helper with a supplied 50 ms response-header bound made the same focused test pass.

This proves the missing containment independently of the unknown field trigger.

## Expected behavior

When an upstream operation stops making progress, the gateway should:

1. terminate the wait after a configurable, documented bound;
2. return a sanitized and accurate gateway timeout response;
3. ensure future requests do not continue leasing the same suspect client/pool generation;
4. avoid automatically replaying requests, especially non-idempotent or streaming requests;
5. preserve long-lived response streaming after response headers have arrived;
6. preserve existing TLS-verification policy, including separation of normal and explicitly configured skip-verification clients.

## Proposed direction for discussion

This is intentionally a proposal, not a requirement for a particular implementation.

One possible implementation is:

- configure a bounded upstream connect timeout;
- bound only the wait for response headers (`RequestBuilder::send()`), not the streamed response body;
- on a response-header timeout, return a sanitized HTTP 504 such as `upstream_timeout`;
- rotate the affected shared `reqwest::Client` generation so subsequent requests lease a fresh pool;
- do not automatically retry/replay the failed request;
- make concurrent timeouts against one generation perform at most one rotation;
- keep normal-verification and skip-verification client generations independent.

Maintainers may prefer a different implementation. The core request is bounded failure plus automatic escape from suspect pooled state, without weakening TLS or replaying requests.

## Suggested regression coverage

- silent upstream accepts TCP/request but never sends headers: bounded timeout;
- timeout maps to sanitized HTTP 504 `upstream_timeout`;
- future request uses a fresh client generation and succeeds;
- concurrent timeouts rotate one generation once;
- normal-verification and skip-verification pools remain independent;
- headers arrive before the bound but the body streams longer: stream succeeds;
- no automatic replay, including POST;
- existing redirect, TLS, credential-injection, and streaming tests remain green.

## Related but distinct reports

- #424 concerns AAAA SERVFAIL/timeout handling and IPv4 fallback.
- #251 concerned Git credential injection on `github.com` and was fixed separately.
- PR #259 concerned absolute-form HTTPS routing and later documented a wrong-port axios case.

None of those appears to track an indefinitely pending upstream request plus client-pool recovery.

SenulMapa:
## Problem

A gateway can enter a state where upstream requests stop completing. The
client-facing TLS handshake succeeds and ALPN negotiates, but no HTTP response
headers ever arrive. The process and listener stay healthy, so recovery
requires restarting the container — interrupting every user of the gateway.

The forwarding path awaited `upstream.send()` with no deadline, so that future
could stay pending indefinitely, and there was no way to escape suspect pooled
state short of a restart.

The initiating cause is **not** proven. A poisoned pooled HTTP/1.1 keep-alive
connection is plausible but unverified. This PR fixes the two deterministic
defects — the missing bound and the missing recovery path — and does not claim
to have identified or fixed the root cause.

## Fix

- **Bounded header wait.** `tokio::time::timeout` around `send()`, which
  resolves at the response head. Configurable via
  `GATEWAY_UPSTREAM_HEADER_TIMEOUT_SECS`, defaulting to 300s.
- **Sanitized failure.** HTTP 504 with the stable identifier
  `upstream_timeout`, marked `x-should-retry: false`. No URLs, credentials, or
  transport detail reach the client.
- **Generation rotation.** Requests lease a `(generation, client)` pair. A
  header timeout retires that generation if it is still current, so later
  requests lease a freshly built client. Concurrent timeouts against one
  generation perform exactly one rotation.

`reqwest`'s own timeout could not be used for the bound: `ClientBuilder::timeout`
and `RequestBuilder::timeout` are total deadlines still enforced while the
response body streams (`TotalTimeoutBody::poll_frame` polls the same timer on
every frame), so either would have capped SSE and long downloads. `read_timeout`
would likewise fire on an idle-but-healthy stream. Bounding `send()` — which
resolves once the response head is read — is what confines the deadline to
headers alone.

The default is deliberately generous: this bound is a containment backstop
against an operation that will never make progress, not a latency budget. The
gateway fronts providers whose non-streaming responses can legitimately hold
headers for minutes, and the same wait also covers writing a large request body
upstream. A tight bound would fail healthy requests *and* churn healthy pools
underneath them. Happy to retune if maintainers prefer a different value.

A separate connect timeout was not added: connecting happens inside `send()`,
so the header bound already subsumes it.

## Safety semantics

- **Requests are NOT automatically replayed.** Not POST, not PUT/PATCH, not
  streaming bodies, not anything. A timeout means the outcome is ambiguous — the
  upstream may already have executed the operation. The gateway reports and
  stops. Rotation only changes which pool the *next* request uses.
- **Response streaming is unaffected.** The deadline covers only the wait for
  headers. Once headers arrive the body streams with no deadline, so SSE and
  long downloads behave exactly as before.
- **In-flight requests are not disturbed by rotation.** Each holds its own
  client clone; the old pool lives until the last one drops it.
- **TLS separation preserved.** The verifying and skip-verification clients are
  separate rotators with separate generations, and each rebuild reproduces the
  posture it was constructed with, so rotation cannot promote one to the other.
- Credential injection, policy enforcement, redirects, approval flows,
  request/response streaming, and the WebSocket path are unchanged. The
  WebSocket `TlsConnector`s are deliberately untouched — they have no pool and
  `reqwest` is not involved.

## Test coverage

12 new tests; `cargo test --workspace` is green at 909 passed / 0 failed.

Through the real forwarding path (`forward_request`, driven over a socket by a
real hyper server):

- a POST to an upstream that accepts and reads but never sends headers returns
  a sanitized 504 `upstream_timeout`, rotates the generation exactly once, and
  reaches the upstream exactly once

This test was verified to fail without the fix: with the bound removed, the
request hangs until the test's own outer guard fires, reproducing the reported
behavior.

At the client/rotator level:

- silent upstream terminates on the configured bound
- headers inside the bound with a body several times slower still streams to
  completion (the SSE / long-download guarantee)
- a timed-out POST is not replayed
- rotation retires the leased generation; a later request leases the
  replacement and completes successfully against a healthy upstream
- concurrent timeouts against one generation rotate exactly once
- a stale generation cannot discard the newer client that replaced it
- verifying and skip-verification pools rotate independently, in both
  directions
- a rebuilt client preserves its original TLS posture
- the default deadline stays a containment backstop

Plus the sanitized 504 response shape: status, `x-should-retry: false`, stable
`error` identifier, and an asserted absence of URLs, credentials and transport
detail in the body.

Existing redirect, TLS, credential-injection and streaming tests remain green.
`cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all
--check`, `RUSTDOCFLAGS="-D warnings" cargo doc` and a release build are all
clean.

## Known limitation

The OAuth token-exchange paths in `inject/` and `apps/` still use bare
`reqwest::Client::new()` with no bound. Same shape, different code path — left
out of scope here since #493 concerns the gateway forwarding path.

Fixes #493


SenulMapa:

guyb1:
great catch on that @SenulMapa 
thank you, will merge it soon!

PR #545, opened 2026-09-02
fix(gateway): bypass credential_not_found for HTML and browser traffic on unknown hosts

### LINKED ISSUE #533: Gateway wrongly requires credential for agent-browser page-text reads on never-before-seen hostnames
**Summary**

A browser-automation tool (`agent-browser`, used from an agent container behind the OneCLI HTTPS gateway) can `open` and fully render a plain public webpage with no credential prompt, but a subsequent text-extraction step on that same already-open page fails with `credential_not_found`, even though the site needs no authentication at all.

**Repro**

1. `agent-browser open https://www.sportskeeda.com/mmo/how-easily-complete-missions-journey-ii-the-division-2-red-horizon-y8s3` — succeeds clean, no credential prompt, page renders.
2. `agent-browser get text body` on that same open page — returns, instead of page text:

```json
{"error":"credential_not_found","hostname":"www.sportskeeda.com",
 "message":"No credentials configured for www.sportskeeda.com in OneCLI...",
 "path":"/mmo/how-easily-complete-missions-journey-ii-the-division-2-red-horizon-y8s3",
 "secret_url":"https://<hive-instance>/w/<workspace>/connections/custom?create=generic&host=www%2Esportskeeda%2Ecom&path=%2F%2A"}
```

Confirmed reproducible on 5 distinct hostnames (sportskeeda.com, screwfix.com, wilko.com, homebase.co.uk, nufarm.com), across 3 separate agent containers/workspaces, 100% hit rate on retest. Common domains that had previously been fetched by these containers (google.com, bbc.co.uk, example.com) pass `get text` clean in the same test runs.

**Why this looks like a gateway bug, not intended behavior**

- `open` (full navigation) on the exact same URL never triggers a credential check.
- The credential gate is documented (gateway skill instructions) as being for actual API calls to connected services (Gmail, GitHub, Stripe, etc.) — not for reading arbitrary public webpages via a browser tool.
- None of the failing domains have any legitimate reason to need a stored credential for a page read, and the response actively invites creating a generic credential connection for them, which would be the wrong fix (masks the bug, grants an unnecessary credential).

**Working theory (unconfirmed — no gateway-internals access to verify further)**

`get text`'s extraction step may route through a lightweight HTTP fetch over the same proxied path used for real API calls, rather than reading Chromium's already-rendered DOM. If so, any *first-time-seen hostname* hitting that path gets treated like an unconnected API integration and gated, regardless of whether the destination needs auth. This would explain why previously-fetched common domains pass while every never-before-seen domain we tried failed identically — but this is a hypothesis based on external behavior only, not a confirmed root cause.

**Impact**

Any agent doing routine web reads via `agent-browser` behind the gateway is likely to trip this on the first new domain it visits, and could be tempted to click the provided credential-connection link to "fix" it — which would grant an unneeded credential rather than address the underlying gating bug.

Searched existing issues for `credential_not_found` and `browser credential` before filing; found none describing this specific interaction between `agent-browser`-style page reads and the credential gate.

Adityakk9031:
Fixes #533.

### Summary
When an agent container behind the HTTPS gateway accesses public websites (e.g. via \gent-browser\, headless Chromium, or scrapers), CDNs/WAFs (like Cloudflare, Akamai) or public servers may return \401 Unauthorized\ or \403 Forbidden\ (such as anti-bot challenge pages, WAF blocks, or forbidden web directories).

Previously, the gateway unconditionally intercepted all 401/403 responses on unknown hosts when no credentials were injected (\injection_count == 0\) and substituted a synthetic JSON \credential_not_found\ error pointing to a custom connection form. This broke browser tools and web text extractors by stripping the upstream HTML response and replacing it with JSON.

### Changes
- Added \is_html_or_browser_traffic\ helper in \orward.rs\ which checks:
  1. If upstream responded with an HTML \Content-Type\ (e.g. \	ext/html\, \pplication/xhtml+xml\).
  2. If client requested HTML via the \Accept\ header.
- Updated 401/403 and auth-400 handling on unknown hosts to bypass \credential_not_found\ and forward the upstream status and headers as-is for HTML/browser web traffic.
- Added unit tests for \is_html_or_browser_traffic\ in \orward.rs\.


Adityakk9031:
@guyb1 have a look

PR #546, opened 2026-09-02
feat(api): allow packaged deployments to supply provider options

### LINKED ISSUE #526: Allow packaged deployments to supply provider options
## Scope

The packaged control-plane composition point is `apps/web/src/lib/api/app.ts`, which is the sole caller of `createApiApp`. The Rust `onecli-gateway` is a separate data-plane process: it performs no control-plane principal resolution and writes no audit records, so it needs no equivalent hook. This proposal is therefore confined to the TypeScript control-plane API.

## Background

OneCLI already supports provider composition. `createApiApp(sessionProvider, options)` accepts `roleResolver`, `resourceHooks`, `connectionHooks`, `crypto`, `oauthOrg`, `appAvailability`, `strictApiKeyAuth` and others, registering each through the existing `init*` / `get*` provider functions. Editions use this today.

## Problem

That capability is available only to code that calls `createApiApp`. In a packaged deployment the application composes itself at `apps/web/src/lib/api/app.ts`, so an operator running the published image cannot supply provider implementations without patching the image.

Deployments that need a deployment-specific provider must therefore fork, even though the provider contract they need already exists and is stable.

## Part 1 — packaged composition

Let the packaged composition module optionally merge provider options supplied by the deployment — for example, by loading a module named in an environment variable and merging its exported options with the built-in ones.

Explicit non-goals: no plugin framework, no new lifecycle, no new provider concepts. This exposes an existing capability to a deployment context that currently cannot reach it. If no module is configured, behaviour is byte-identical to today.

## Part 2 — non-human principals

OneCLI currently models an authenticated API-key caller as a `User`: `ApiKey` carries `userId` and `userEmail`, `ensureApiKey` and `regenerateApiKey` key off `userId`, and `RoleResolver` is `getUserRole(userId, organizationId)`.

A deployment needing a service or machine caller must therefore either borrow a human credential, fabricate a `User` row, or use an organization-scoped bootstrap key. Each of these either misattributes the actor in audit or grants far more authority than the caller needs.

Proposed: a principal-shaped provider interface alongside `RoleResolver`, returning a small generic `PrincipalContext` of `principalId`, `principalType` and an opaque `authorizationContext`.

Upstream would own identity resolution and composition. Deployment-specific authorization models stay behind that seam and are explicitly **not** proposed for upstream.

## Part 3 — audit attribution

`AuditLog.userId` is currently `NOT NULL` with a required foreign key to `User`, so a non-human actor cannot be recorded without fabricating a user.

Proposed minimum: add a nullable `principalId` and an `actorType`; keep `userId` populated for human actors; allow it to be null for non-human actors.

`AuditLog` has one write site (`audit-service.ts`, `withAudit`) and no read sites, so this is a single chokepoint rather than a diffuse change.

## Backward compatibility

Human sessions and personal API keys are unchanged: `userId` stays populated for human actors, every new field is additive or a relaxation, and Part 1 is inert unless a deployment configures it.

Nothing in this proposal requires any deployment-specific package to be installed for existing behaviour to work.

## Motivating examples

An orchestrator process that provisions and configures agents needs to authenticate as itself, hold only the operations it actually calls, and appear in audit as a machine rather than as whichever engineer's key it borrowed.

The same shape applies to CI runners, schedulers, and other automation that acts on its own behalf rather than on behalf of a person.

---

Happy to discuss scope or shape before any code — raising this as an issue first per `CONTRIBUTING.md`. If the direction is acceptable, we're willing to contribute the implementation.

Adityakk9031:
Fixes #526.

### Summary
Allows packaged deployments (such as container images) to supply custom provider options without requiring a source fork.

### Changes
1. **Provider Options in \createApiApp\**:
   - Expanded \CreateApiAppOptions\ in \packages/api/src/app.ts\ to accept all provider interfaces (\crypto\, \sshCa\, \eventBus\, \ttachmentStore\, \oauthOrg\, \strictApiKeyAuth\, \workspaceAccessChecker\, \sessionEnforcer\, \sessionThrottle\, \
esourceHooks\, \connectionHooks\).
   - Added \initConnectionHooks\ to \connection-hooks.ts\ and wired all provider slots inside \createApiApp\.
2. **Packaged Composition Module Loader**:
   - Added \loadDeploymentProviderOptions\ in \pps/api-server/src/app.ts\ to optionally import custom options from a module path configured via the \ONECLI_PROVIDER_OPTIONS_MODULE\ environment variable and merge them into \createApiApp\.
3. **Tests**:
   - Added unit tests in \packages/api/src/app-provider-options.test.ts\ verifying that custom provider options are correctly registered.


Adityakk9031:
@guyb1 have a look

PR #552, opened 2026-09-08
fix(api): atomically consume CLI auth poll session

siddiqueirshad:

YES

Bug fix

During CLI device auth, the install script polls `GET /v1/auth/cli/poll?code=...` until the browser confirms the session. After confirmation, `pollCliAuthSession` reads a `confirmed` row with a plaintext `apiKey`, returns it to the caller, and then clears the row.

That read-then-update is not atomic. Two concurrent poll requests can both observe `status=confirmed` with a non-null `apiKey` and both return the workspace API key before either update lands. The code comment already claims the key is consumed once, and `confirmCliAuthSession` uses a CAS-style update, but the poll path did not.

## What is the new behavior?

Poll now consumes the session with an atomic `updateMany` guarded on `code`, `status: confirmed`, and `apiKey IS NOT NULL`, matching the CAS pattern used by `confirmCliAuthSession`. Only the poll that wins the update returns `status: ok` with the API key; the loser gets `status: expired`. Already-consumed sessions also return `expired`.

### Root cause

Non-atomic read-then-update in `packages/api/src/services/cli-auth-service.ts` (`pollCliAuthSession`).

### Issue

Independently discovered through code review and execution-path analysis. I did not find an existing upstream issue or open PR for this race.

### Testing

- `pnpm --filter @onecli/api test src/services/cli-auth-service.test.ts` — 3 passed
- `pnpm --filter @onecli/api test src/services/cli-auth-service.pg.test.ts` — skipped locally (no `POLICY_PROOF_DATABASE_URL`); runs in CI against real Postgres
- `pnpm --filter @onecli/api lint` on changed files — passed

### Reproduction (conceptual)

1. Create a CLI auth session and confirm it in the browser.
2. Fire two concurrent `GET /v1/auth/cli/poll?code=...` requests.
3. Before this fix, both can return `{ status: ok, api_key: ... }`.
4. After this fix, exactly one returns `ok`; the other returns `expired`.

### DIFF
--- a/packages/api/src/services/cli-auth-service.pg.test.ts
+++ b/packages/api/src/services/cli-auth-service.pg.test.ts
@@ -1,0 +1,103 @@
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { randomBytes } from "node:crypto";
+
+import { proofDatabaseUrl } from "../testing/pg-proof.js";
+
+/**
+ * CLI auth poll must hand out the workspace API key at most once. The unit
+ * suite mocks the CAS boundary; this suite proves the race on real Postgres.
+ */
+
+const PROOF_URL = proofDatabaseUrl();
+
+type Db = typeof import("@onecli/db").db;
+type CliAuth = typeof import("./cli-auth-service");
+
+let db: Db;
+let pollCliAuthSession: CliAuth["pollCliAuthSession"];
+
+const P = "cli-auth-";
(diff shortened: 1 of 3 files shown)


siddiqueirshad:

siddiqueirshad:
Hi @breken-ai, when you get a chance, could you please take a look at this PR? Thanks!

PR #559, opened 2026-09-11
fix(web): preserve API status for malformed error bodies

### LINKED ISSUE #558: API client throws a TypeError instead of ApiError when an error body is JSON null
### Description

`refusal()` in `apps/web/src/lib/api/client.ts` turns every non-OK response from the JSON helpers (`apiGet`, `apiPost`, `apiPatch`, `apiPut`, `apiDelete`) and `uploadImage` into an `ApiError` carrying the HTTP status. Callers branch on that status; for example, `direct-thread-section.tsx` shows a 409 inline.

```ts
const extractErrorMessage = (body: Record<string, unknown>, status: number) => {
  const err = body.error;
  if (typeof err === "string") return err;
  if (err && typeof err === "object" && "message" in err)
    return String((err as { message: unknown }).message);
  return `Request failed: ${status}`;
};

export const refusal = async (res: Response): Promise<ApiError> => {
  const body = await res.json().catch(() => ({}));
  return new ApiError(extractErrorMessage(body, res.status), res.status);
};
```

The `.catch` handles bodies that don't parse, but nothing checks what a successful parse returned:

- **A body of `null`** parses fine, then `body.error` throws `TypeError: Cannot read properties of null (reading 'error')`. The caller gets that `TypeError`, which has no `status`, instead of an `ApiError`.
- **An `error.message` object without a callable `toString`**, such as `{"error":{"message":{"toString":null}}}`, makes `String(...)` throw a `TypeError` too.
- **Other non-string messages** don't throw, but reach the UI as `"42"`, `"null"` or `"[object Object]"`.

`uploadAttachment` and `fetchAttachmentBlob` in `lib/api/attachments.ts` use a private copy of the same parse (`refused`). It already handles a `null` body, but has the same `String(...)` coercion.

The API's own error handler always sends a string `error.message`. This only happens when something else answers the request with an unexpected JSON body, a proxy for instance. When it does, status-specific handling is skipped.

### Steps to reproduce

1. `await refusal(new Response("null", { status: 403 }))` rejects with a `TypeError` instead of resolving to an `ApiError` with `status: 403`.
2. With `apiFetch` returning `new Response("null", { status: 409 })`, `apiPost(...)` rejects with the same `TypeError`, so `error.status` is `undefined`.
3. `await refusal(new Response('{"error":{"message":{"toString":null}}}', { status: 409 }))` also rejects with a `TypeError`.

### Expected behavior

A failed response always becomes an `ApiError` with its status. A string `error`, or a string `error.message`, becomes the message, and any other shape falls back to `Request failed: <status>`.

I have a small fix with a regression test and will open a PR for it.

breken-ai:

YES

Bug fix, with a regression test.

Fixes #558

`refusal()` passes whatever `res.json()` parsed straight to `extractErrorMessage`, which reads `body.error` and calls `String()` on `error.message`. A `null` body, or a message object whose `toString` isn't callable, throws a `TypeError` before the `ApiError` is built. Callers of the JSON helpers and `uploadImage` then lose the HTTP status. `attachments.ts` has its own copy (`refused`), which already handles a `null` body but has the same `String()` coercion, so `uploadAttachment` and `fetchAttachmentBlob` still throw on the `toString` case.

The API's own error handler always sends string messages, so this only happens when something in between, such as a proxy, answers with an unexpected JSON body.

## What is the new behavior?

`extractErrorMessage` now takes `unknown`. It uses the message only when `error` is a string or `error.message` is a string; anything else falls back to the existing `Request failed: <status>`. `refusal()` therefore always returns an `ApiError` with the response status.

`attachments.ts` now uses `refusal` instead of its own copy, matching the comment on `refusal` that every refusal parse lives there.

One intentional difference: a non-string `error.message` (a number, boolean, object, array or `null`) now gets the fallback text instead of coerced text such as `"42"` or `"[object Object]"`. For example, `plan-switch-dialog.tsx` shows the server's message unless it starts with `Request failed`, so for those bodies it now shows its own generic copy. Empty-string messages behave as before.

`lib/cli-auth/api.ts` parses its own errors into a plain `Error` with different fallback copy. I left it alone.

## Additional context

`apps/web/src/lib/api/client.test.ts` covers `refusal` with:
- a `null` body
- an uncallable `toString` message
- a numeric message
- the supported string envelopes, including an empty message

It also checks that `apiPost` rejects with the status when the error body is `null`.

Checked locally on Node 22.22.0 with pnpm 9.0.0, sharing a turbo cache with a clean run on `main`:

| Check | Result |
|---|---|
| New test against unpatched `main` (`3e595ef`) | 4 failed, 3 passed (7 tests) |
| New test with this change | 7 passed (7 tests) |
| `pnpm --filter @onecli/web test` | 871 passed, 8 skipped (879 tests); on `main`, 864 passed, 8 skipped (872 tests) |
| `pnpm build` | passed |
| `pnpm check` | passed |

Not tested in a browser against a running instance.

### DIFF
--- a/apps/web/src/lib/api/attachments.ts
+++ b/apps/web/src/lib/api/attachments.ts
@@ -1,5 +1,5 @@
 import { apiUpload, apiFetch } from "@/lib/api-fetch";
-import { ApiError } from "./client";
+import { refusal } from "./client";
 import { conversationPath } from "./conversations";
 import type { AttachmentMeta } from "./types";
 
@@ -9,27 +9,6 @@ import type { AttachmentMeta } from "./types";
  * Content-Type. Metadata comes back as ordinary JSON; bytes come back as a
  * Blob the caller turns into an object URL.
  */
-
-/**
- * The house error envelope is `{ error: { message, type } }` for every
- * ServiceError-mapped refusal, and a bare `{ error: "…" }` for the handful of
- * routes that answer directly (the upload's 413). Read BOTH — stringifying
- * the object rendered every refusal as "[object Object]".
- */
-const refused = async (res: Response): Promise<ApiError> => {
-  const body: unknown = await res.json().catch(() => ({}));
-  const err =
-    body && typeof body === "object" && "error" in body
-      ? (body as { error: unknown }).error
-      : undefined;
-  const message =
-    typeof err === "string"
-      ? err
-      : err && typeof err === "object" && "message" in err
-        ? String((err as { message: unknown }).message)
-        : `Request failed: ${res.status}`;
-  return new ApiError(message, res.status);
-};
 
 export const uploadAttachment = async (
   conversationId: string,
@@ -44,7 +23,7 @@ export const uploadAttachment = async (
     contentType: file.type || "application/octet-stream",
     signal,
   });
-  if (!res.ok) throw await refused(res);
+  if (!res.ok) throw await refusal(res);
   return (await res.json()) as AttachmentMeta;
 };
 
@@ -59,6 +38,6 @@ export const fetchAttachmentBlob = async (
       `/attachments/${encodeURIComponent(attachmentId)}`,
     ),
   );
-  if (!res.ok) throw await refused(res);
+  if (!res.ok) throw await refusal(res);
   return res.blob();
 };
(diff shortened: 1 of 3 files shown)


breken-ai:

guyb1:
@breken-ai great catch and thanks for the PR!
merging

